PR #276, opened 2012-06-04
Store the offline manifest entries in the cache.

mpessas:
We make the `compress` command store the entries of the manifest.json file in cache and the `compress` templatetag try to read the correct value from the cache instead of the `manifest.json` file.


diox:
Closing this because it is super-old and I'm really not a fan of adding more layers of complexity to offline caching anyway.

Issue #30, opened 2011-04-15
gather, concatenate and *then* pass to filters?


simonwiles:
Hi!  Is there anyway to collect and concatenate the CSS or JS in a particular block and _then_ pass it through a filter chain?  In particular, I'd like to use pyScss to process my SCSS.  I've created a `pySCSSFilter` extending `FilterBase`, and I'd really like to do something like:

```
{% compress css %}
<link rel="stylesheet" type="text/css" media="screen" href="{{ MEDIA_URL }}css/mixins.scss" />
<link rel="stylesheet" type="text/css" media="screen" href="{{ MEDIA_URL }}css/base.scss" />
<link rel="stylesheet" type="text/css" media="screen" href="{{ MEDIA_URL }}css/other.scss" />
{% endcompress %}
```

but of course doesn't won't work as my files are processed in isolation from one another.

Is there a way?

Cheers!

k-funk:
@airtonix I started doing this, but the changes to any `@import`'ed files don't get watched, and so the server needs to be restarted for every css change in them.
Kinda amazed this feature hasn't been implemented yet. [webassets](https://webassets.readthedocs.io/en/latest/builtin_filters.html#sass) solved this, allowing users to use `SASS_AS_OUTPUT = True`, which optionally concats the files first.

PR #289, opened 2012-07-17
Adding option to concatenate files before processing

jannon:
### What's in this pull request
- Let tag arguments after kind (css|js) be specified in any order 
- Use more standard 'as <name>' syntax to name the code block
- Add ability to specify options in the tag
- 'group_first' option to allow files to be concatenated before processing

There are tests and documentation updates as well as the code changes
#### Tag Argument Order

There's not really any need to enforce a certain arrangement of arguments in the tag so let's just let people specify them however they want after the kind parameter.  So now all of these are valid:

```
{% compress js inline as main %}

{% compress js as main file %}

{% compress css group_first=true as base %}
```
#### Using 'as &lt;name>'

Lots of things in python and django use this syntax, so it would be nice to do so as well
#### Specifying options in the tag

Just like the `with` tag in django, you can specify an arbitrary number of keyword/value arguments to the tag.  These will get put into a dictionary and passed to the Compressor classes.  Also, the default compressors, add these to the options that they pass the precompilers and filters.  One of the main motivations for this was to prepare for being able to specify that files should be grouped first before, say, lessc precompilation, which leads to...
#### 'group_first' option

Setting the option `group_first=true` in the `compress` template tag will cause files of the same type to be concatenated before being handed to precompilers/filters.  This will allow people to do standard code decomposition with things like .less files (having a mixins file separate from page-specific files)

For example:

```
{% compress css group_first=true %}
    <link rel="stylesheet" href="/media/css/mixins.less" type="text/less" />
    <link rel="stylesheet" href="/media/css/home.less" type="text/less" />
    <link rel="stylesheet" href="/media/css/foo.css" type="text/css" />
{% endcompress %}
```

Will concatenate the .less files so the less precompiler will properly process all the files
This provides a fix for #30

### DIFF
--- a/compressor/templatetags/compress.py
+++ b/compressor/templatetags/compress.py
@@ -116,7 +116,7 @@ class CompressorMixin(object):
                 raise
 
         # Or don't do anything in production
-        return self.render_result(self.get_original_content(context), context, deferred)
+        return self.render_result(self.get_original_content(context), context, deferred, mode)
 
     def render_output(self, compressor, mode, forced=False):
         return compressor.output(mode, forced=forced)


k-funk:
Why was this never merged? :(

karyon:
@k-funk this has been continued in #294. updating that PR would be greatly appreciated :)

PR #294, opened 2012-07-30
Support for modern webdev practices/technologies

### LINKED ISSUE #30: gather, concatenate and *then* pass to filters?

### LINKED ISSUE #249: Support script loaders: Add a template tag for getting the url of a compressed file
Django 1.4 adds the [static](https://docs.djangoproject.com/en/1.4/ref/contrib/staticfiles/#template-tags) template tag. However, this still produces the original URL of the specified file, rather than the compressed version.

This isn't an issue using the normal `{% compress … %}` tag with `<script src="…">` elements. However, if one is using script loader (like LAB.js), there's no obvious way to get the compressed URL of a javascript file for the loader to reference.

This could be addressed by a template tag that compresses the given file and returns the compressed URL. For example:

```
<script>$LAB.script('{% compress_url "js/foo.js" %}')</script>
```

would produce:

```
<script>$LAB.script('/static/CACHE/js/abcdef.js')</script>
```

Thanks for considering this.

jannon:
Okay here is a pull request that repackages and cleans up a couple earlier requests
### What's in this pull request
- Let tag arguments after kind (css|js) be specified in any order 
- Use more standard 'as <name>' syntax to name the code block
- Add ability to specify options in the tag
- Ability to group files before processing
- Ability to redirect tag output to template context variable

There are tests and documentation updates as well as the code changes
#### Tag Argument Order

There's not really any need to enforce a certain arrangement of arguments in the tag so let's just let people specify them however they want after the kind parameter.  So now all of these are valid:

```
{% compress js inline as main %}

{% compress js as main file %}

{% compress css group_first=true as base %}
```
#### Using 'as &lt;name>'

Lots of things in python and django use this syntax, so it would be nice to do so as well
#### Specifying options in the tag

Just like the `with` tag in django, you can specify an arbitrary number of keyword/value arguments to the tag.  These will get put into a dictionary and passed to the Compressor classes.  Also, the default compressors add these to the options that they pass the precompilers and filters.  One of the main motivations for this was to prepare for being able to specify that files should be grouped first before, say, lessc precompilation (as discussed in #30)  Currently supported options:
- group_first - if 'group_first=true' is specified, files of like type will be concatenated before being handed to precompilers/filters
- deferred - if 'deferred=true' is specified AND 'as <name>' is specified the results of the template tag are stored in a <name> template context variable
#### Concatenating files before processing (fixes #30)

As mentioned above, using the group_first=true tag options will concatenate files before processing.   This is useful for having django_compressor work for best practices like separate mixins.less and page.less files
#### Template context variable (fixes #249)

As mentioned above, using 'deferred=true' and 'as <name>" in the template tag will cause the output to be placed in a template context variable rather than rendered immediately. This is useful for capturing the results of the compression (the url) for use with asynchronous resource loaders (e.g. yepnope.js, require.js, etc.).  In fact, if the output mode is 'file', the 'deferred' option set the template variable to be just the compressed url (without 'script' tag).  If the mode is 'inline' the template variable just contains whatever the output would have been normally

### DIFF
--- a/compressor/base.py
+++ b/compressor/base.py
@@ -31,7 +31,7 @@ class Compressor(object):
     """
     type = None
 
-    def __init__(self, content=None, output_prefix=None, context=None, *args, **kwargs):
+    def __init__(self, content=None, output_prefix=None, context=None, opts=None, *args, **kwargs):
         self.content = content or ""
         self.output_prefix = output_prefix or "compressed"
         self.output_dir = settings.COMPRESS_OUTPUT_DIR.strip('/')
@@ -42,6 +42,7 @@ class Compressor(object):
         self.extra_context = {}
         self.all_mimetypes = dict(settings.COMPRESS_PRECOMPILERS)
         self.finders = staticfiles.finders
+        self.opts = opts or {}
 
     def split_contents(self):
         """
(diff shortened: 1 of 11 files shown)


k-funk:
Would really like to see this go out if you still have time @jannon . If not, let us know and hopefully someone else can work on it.

iamareebjamal:
If a PR with conflicts resolved and passing tests is made, will it be merged?

Wanting to ensure my time and efforts won't go to waste

karyon:
generally, sure, if the code looks good and the benefit justifies the added complexity. Although I'd prefer to have at least separate commits if not separate PRs for the individual features, so they can be reviewed/merged separately. otherwise it might very well stall again as this PR did :)

Is there anything in particular you're interested in? I'm wondering e.g. whether loaders (which seem to be the motivation for "Template context variable") are still a thing...

iamareebjamal:
I'm mainly interested in merging resources before passing them to the processor.

So, yeah, I'll create a focused PR on that and move forward with one PR for one feature (unless the requirements of features overlap the code changes)

PR #331, opened 2012-10-31
Apply CssAbsoluteFilter to generated CSS even when COMPRESS_ENABLED is False

greghball:
When COMPRESS_ENABLED is False, and applying pre-compilers to produce CS......S, apply the CssAbsoluteFilter input method if that filter is configured in settings.COMPRESS_CSS_FILTERS.

Without this filter, any urls with relative paths in the LESS/SASS/SCSS etc. source will be broken in the output, making pre-compilation useless.

### DIFF
--- a/compressor/base.py
+++ b/compressor/base.py
@@ -16,6 +16,7 @@ from compressor.conf import settings
 from compressor.exceptions import (CompressorError, UncompressableFileError,
         FilterDoesNotExist)
 from compressor.filters import CompilerFilter
+from compressor.filters.css_default import CssAbsoluteFilter
 from compressor.storage import default_storage, compressor_file_storage
 from compressor.signals import post_compress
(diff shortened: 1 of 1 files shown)


karyon:
@diox, do you still think a setting is required? imo this is not magical and arbitrary at all. when django-compressor changes file locations (e.g. by precompiling them), it should take care that file paths inside the moved files stay valid.

although i'd do that a bit differently, maybe give filter a param `filters` which is self.cached_filters for the original two callers and CssAbsoluteFilter for the added call.

what do you say?

diox:
@karyon : can you submit a pull request with that idea ? I've warmed up to the idea of not having an extra setting ; I'm still not quite sure how the right way would look like, but a new pull request would be helpful to restart the discussion.

karyon:
yes, will do, hopefully later today.

diox:
Closing in favor of #653.

PR #346, opened 2012-12-14
Add caching to CompilerFilter to not re-compile unchanged files

pindia:
Currently django_compressor seems to re-run precompilers on each compression, even if files are completely unchanged. I haven't been able to stop this through any combination of existing caching options. In my large project, this has gotten to the point where it takes 10-15 seconds to load every page. 

This patch adds caching directly to the `CompilerFilter` class. Before running the compiler, it computes a hash of the input contents and checks the compressor cache for a cached output. If not found, the compiler runs as normal and the output is saved to the cache for next time. Thus, only files that have actually changed are recompiled. When used in combination with a file-based caching backend to persist across server restarts, this greatly speeds up development.

### DIFF
--- a/compressor/base.py
+++ b/compressor/base.py
@@ -15,7 +15,7 @@ from compressor.cache import get_hexdigest, get_mtime
 from compressor.conf import settings
 from compressor.exceptions import (CompressorError, UncompressableFileError,
         FilterDoesNotExist)
-from compressor.filters import CompilerFilter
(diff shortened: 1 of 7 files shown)


staab:
:+1: on that solution, ejo. Does exactly what I want.

diox:
Fixed in #650, which includes a rebase of this PR. Thanks!

PR #353, opened 2013-01-05
Use multiprocessing to speed up offline compression

pindia:
For large projects, offline compression starts to take a while because there are many templates with many files each to process. But since each template is compiled separately, we can use multiprocessing to compile them at the same time.

This patch adds an option `--processes` to the `compress` manager command to instruct it to use multiple processes.

I'm submitting this request before writing tests and documentation for the new behavior to see if this is something that you'd be interested in merging. If so, I'm happy to do both.

### DIFF
--- a/compressor/management/commands/compress.py
+++ b/compressor/management/commands/compress.py
@@ -105,7 +105,7 @@ def patched_get_parent(self, context):
     return compiled_template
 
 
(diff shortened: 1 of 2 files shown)


karyon:
superseded by #737 which looks nicer and is less out of date :)

PR #362, opened 2013-02-05
Fix for double render in sekizai postprocessor (fixes #361)

### LINKED ISSUE #361: sekizai postprocessor reparses the Template
This seems to be a better solution for the sekizai postprocessor.

The data passed in has already been rendered, so no need to create a Template object. By passing it in to Template, any swig templates in javascript (or text created by {% templatetag %}) will get parsed as django template source, which is probably not intended.

    from compressor.templatetags.compress import CompressorNode
    from django.template.base import TextNode

    def compress(context, data, name):
        """
        Data is the string from the template (the list of js files in this case)
        Name is either 'js' or 'css' (the sekizai namespace)
        Basically passes the string through the {% compress 'js' %} template tag
        """
        return CompressorNode(nodelist=TextNode(data), kind=name, mode='file').render(context=context)

aarcro:
Here's a pull request for the issue I opened. 

(I'm fairly new at git and github, so hope I'm going about this the right way)

### DIFF
--- a/compressor/contrib/sekizai.py
+++ b/compressor/contrib/sekizai.py
@@ -6,7 +6,7 @@
  and: https://github.com/ojii/django-sekizai.git@0.5
 """
 from compressor.templatetags.compress import CompressorNode
-from django.template.base import Template
+from django.template.base import TextNode
 
 
 def compress(context, data, name):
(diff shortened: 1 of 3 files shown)


blueyed:
See #623 for a follow-up.

PR #411, opened 2013-05-31
Compass suport

ypopovych:
Added support of compass config files.
Realization quick and dirty, but works. Main idea - change command directory to current app directory, so we can put config.rb to this dir.

So we can add to compress precompilers:
('text/x-sass', 'sass --compass -C {infile} {outfile}'),

And then put config.rb to app dir(if we need compass in this app) or in project directory if we need compass for project static folder


diox:
Hi,

This is a good idea, but I think it's a little too much to have in compressor core. You can easily have this live outside compressor by making your own filter, so I'm going to close this.

PR #500, opened 2014-03-31
Allow for more flexible compressor setups

nagyv:
The above is necessary if someone uses django-compressor with django-sekizai, and has other blocks than "js" and "css". It's 100% backward compatible.

### DIFF
--- a/compressor/templatetags/compress.py
+++ b/compressor/templatetags/compress.py
@@ -22,10 +22,11 @@ class CompressorMixin(object):
 
     @property
     def compressors(self):
-        return {
+        defaults = {
             'js': settings.COMPRESS_JS_COMPRESSOR,
             'css': settings.COMPRESS_CSS_COMPRESSOR,
         }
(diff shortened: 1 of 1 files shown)


diox:
Since there has been no reply, and no one else has asked for this and this can easily be done by having your own `compress` template tag, going to close it.

altryne:
@diox sorry to reply to an older then a year thread o_0 
But I'm stumbling on the same issue currently, and I'm not sure what you mean by
`this can easily be done by having your own compress template tag`

Can you share a link on an example? 
I currently have this exact situation, and the ability to render different render blocks with sekizai would be a very helpful one

diox:
I don't have the time to come up with a working example, but it should be straightforward: You need to register your own template tag / sekizai "tag" that inherits from compressor's, and use that tag in your templates.

Make sure your custom tag does use a class inheriting from `CompressorNode` if you want offline compression to work.

altryne:
Thanks for the fast response @diox

Issue #425, opened 2013-07-23
Files to be precompiled first served from STATIC_ROOT instead of STATICFILES_DIRS


croby:
This appears to be related to #85, and this commit: https://github.com/jezdez/django_compressor/commit/02809e396bcce18f340c7f4759e065efe94122f5

After running `collectstatic`, any `.scss` files that have been included are moved to `/static` (`STATIC_ROOT`). When this happens, no edits to the `.scss` file are picked up, as the compressor finds that file first before looking at the app directory where this file exists.

If I move (in `compressor/base.py`)

```python
filename = self.finders.find(urllib.url2pathname(basename))
```

to the top of the `try/except` block, it works as I'd like it to. The commit referenced above inverted this behavior, causing the issue to manifest as I see it.

I'm completely open to the fact that I might not have my settings set up properly, and it seems as though (through my reading of the code in `get_filename` if I didn't have any storages set up, it would work properly. That doesn't seem to work for me, as serving up files seems to bomb right away. When I set a breakpoint in `get_filename`, this is what I see.

```python
(Pdb) print self.storage.path(basename)
/home/vagrant/web/static/css/default.scss

(Pdb) print compressor_file_storage.path(basename)
/home/vagrant/web/static/css/default.scss

(Pdb) print self.finders.find(urllib.url2pathname(basename))
/home/vagrant/web/src/static/css/default.scss
```

I want it to be served from the last location, but it would never have a chance to get there.

diox:
Fixed by #658

PR #505, opened 2014-04-03
In debug mode, use the app's static dir first

abompard:
In debug mode, files should always be compressed based on the app's static directory, not the global static directory, where files can be outdated.

This is related to #425, I found out about this bug after having written this commit. However, my commit includes a unit test (which was much more work than the actual change in the code), so I don't consider my efforts wasted ;-)

### DIFF
--- a/compressor/base.py
+++ b/compressor/base.py
@@ -114,18 +114,19 @@ class Compressor(object):
         get_filename('css/one.css') -> '/full/path/to/static/css/one.css'
         """
         filename = None
-        # first try finding the file in the root
-        try:
-            # call path first so remote storages don't make it to exists,
-            # which would cause network I/O
-            filename = self.storage.path(basename)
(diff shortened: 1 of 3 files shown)


karyon:
@diox this is another one that i stumbled upon while setting up django-compressor in my project. please tell me which of the two approaches (https://github.com/roverdotcom/django_compressor/commit/70dddbe25c0e05e6df7f01eae4b19bcf1eaf1e2c and https://github.com/hyperkitty/django-compressor/commit/2087f977370667631a221a39ca65784e58cdffce) you prefer and whether you need any other changes and i'll try to make a rebase.

diox:
The latter seems better to me. Might just need a little more comments to understand what's going on, it's tricky if you are not super familiar with the problem, took me a minute to figure it out :)

diox:
Fixed by #658

PR #519, opened 2014-05-13
Empty static files cause a missed hit when running in offline mode - check for None rather than False

stefanocrosta:
When verifying the content of cached_offline check for None rather than simply False

An empty file results in an empty string, causes the cache hit to fail and thus to recompress the section.


stefanocrosta:
ups sorry I never replied to that I probably didn't notice your answer at the time :( happy to see it got fixed!

Issue #543, opened 2014-07-28
Using protocol indepdendant STATIC_URL and the default CssAbsoluteFilter break CssDataUriFilter


ntucker:
By using // instead of http:// or https:// for the STATIC_URL, the absolute filter will cause a bunch of urls to start with //. This will then make the dataurifilter ignore those and not convert them.

Simply putting the dataurifilter before will not work because dataurifilter isn't as smart about finding the files as the absolutefilter is.

karyon:
as stated in #545, this was introduced in https://github.com/django-compressor/django-compressor/commit/62c805b26c413224a3b766965e8a5f626d6e86b5 to fix #455.

Issue #455, opened 2013-12-05
URLs in imported CSS


shacker:
Our site uses Zurb Foundation and our base template includes all of Foundation CSS. Foundation 5's _global.scss includes this external reference:

```
@import url("//fonts.googleapis.com/css?family=Open+Sans:300italic,400italic,700italic,400,300,700");
```

Compressor complains:

```
CommandError: An error occured during rendering /foo/templates/page_base.html: [Errno 2] No such file or directory: '/foo/static/fonts.googleapis.com/css'
```

which is because the base template requests compression of a CSS doc it can't handle.

I'm not clear whether this a compressor bug or a matter of configuration on my end.

PR #545, opened 2014-07-30
DataURI doesn't work if using protocol indepdendant static url

ntucker:
fixes https://github.com/django-compressor/django-compressor/issues/543


diox:
Hi,

This was added on purpose in 62c805b26c413224a3b766965e8a5f626d6e86b5 to fix #455. Evidently that causes a problem for you, but I am not sure this is the right solution... Need to investigate how to reconcile both use cases.

karyon:
since this breaks #455 again, closing. #543 stays open of course.

PR #562, opened 2014-09-17
Allow compressor to fallback to remote files

thenewguy:
Attempt to find files remotely when they cannot be found locally if the compressor storage is remote and the 'COMPRESS_FETCH_FROM_REMOTE' setting is True

This includes the Windows compatibility patch from https://github.com/django-compressor/django-compressor/pull/561


certifiedloud:
+1 this project needs this feature.

karyon:
if anyone wants to pick this up, this needs a rebase, tests and documentation.

PR #570, opened 2014-10-15
Resolve conflicts of CSSO support

vovanbo:
Fix of #538


karyon:
hi! we decided that we won't add new filters to keep the maintenance effort as low as possible, as they're quite easy to manually add. in fact, we even thought about removing the existing ones... so i'm closing this, sorry!

PR #572, opened 2014-10-23
Add support for multiple offline contexts.

pope1ni:
Offline compression now supports generation for multiple contexts by
providing a list or tuple of dictionaries, or by providing a dotted
string pointing to a generator function.

This makes it easier to generate multiple contexts dynamically for
situations where a user might be able to select or be served a different
theme for a website, etc.

### DIFF
--- a/compressor/management/commands/compress.py
+++ b/compressor/management/commands/compress.py
@@ -22,6 +22,7 @@ from compressor.conf import settings
 from compressor.exceptions import (OfflineGenerationError, TemplateSyntaxError,
                                    TemplateDoesNotExist)
 from compressor.templatetags.compress import CompressorNode
+from compressor.utils import get_mod_func
 
 if six.PY3:
     # there is an 'io' module in python 2.6+, but io.StringIO does not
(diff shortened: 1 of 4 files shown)


pope1ni:
@diox: Is it possible for this to be merged before the next release?

pope1ni:
I rebased this. The build only failed because of changes to coffin which is now at v2.0.0 and strips out a lot of old stuff...

pope1ni:
Hi @diox 

Any chance of getting this merged? I've rebased it again.

Thanks.

compressor/management/commands/compress.py
```diff
@@ -211,44 +211,54 @@ def compress(self, log=None, **options):
                       "\n\t".join((t.template_name
                                    for t in compressor_nodes.keys())) + "\n")
 
+        contexts = settings.COMPRESS_OFFLINE_CONTEXT
+        if isinstance(contexts, six.string_types):
+            module, function = settings.COMPRESS_OFFLINE_CONTEXT.rsplit('.', 1)
```
diox: I'm not sure if it's much better than your version, but we have a `get_mod_func` in `compressor.utils` that does this.

compressor/management/commands/compress.py
```diff
@@ -211,44 +211,54 @@ def compress(self, log=None, **options):
...
+            contexts = getattr(import_module(module), function)()
```
diox: We should probably catch errors and re-raise as an `ImportError` with a message containing `settings.COMPRESS_OFFLINE_CONTEXT`, similar to how it's done in `compressor.cache` (see `get_cachekey`)

compressor/management/commands/compress.py
pope1ni: Makes sense to use the same approach. The `ValueError` is also caught in `get_mod_func`, although it will lead to an `AttributeError` later due to the empty string for the function name.

compressor/management/commands/compress.py
pope1ni: Looking at how it is done in `compressor.cache`, the `AttributeError` is caught anyway. So updating to the same approach should handle all of the error cases.

pope1ni:
Updated after comments. (When ready I can squash the commits.)

diox:
Can you add a couple tests for the error handling that you just added ? Then you can rebase, it looks good.

pope1ni:
Added extra tests as requested and squashed.

diox:
That looks good, but I just noticed something: do you think you could change the way you override `COMPRESS_OFFLINE_CONTEXT` in tests to use django's `override_settings()` context manager ?

Sorry for not noticing it sooner. It's not a terribly huge deal, but we're trying to move away from the old hacks regarding settings modifications.

diox:
Actually, @karyon changed my mind on this, since all the other tests in that file do it like that, let's keep it and clean it up in a later commit.

diox:
Thanks again for your PR.

PR #576, opened 2014-11-13
Quote filenames in case they contain e.g. spaces, etc.

MikeAmy:
I got an Error: ENOENT when compressing a less CSS file through less because it had a space in the filename. 

Solution: Quote filenames in case they contain e.g. spaces, etc.

Seem to remember contributing a pull request before for this bug, not sure what happened to it.


karyon:
previous version of this PR is here https://github.com/django-compressor/django-compressor/pull/536 and got merged.

PR #577, opened 2014-11-21
Support {% else %} for when compression is off.

mbertheau:
The use case is: a frontend developer uses django to render the templates to the browser. django-compressor doesn't detect changes in source less or sass files and therefore doesn't know when to invalidate the cache. The frontend developer uses grunt or gulp or some similar tool to compile her less or sass files to CSS, which are then served directly by django without involving django-compressor.

### DIFF
--- a/compressor/templatetags/compress.py
+++ b/compressor/templatetags/compress.py
@@ -123,13 +123,16 @@ class CompressorMixin(object):
 
 class CompressorNode(CompressorMixin, template.Node):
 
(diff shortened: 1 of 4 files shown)


diox:
Hi,

Thanks for submitting this and sorry it took that long (almost a year!) to get back to you. I just don't have much time to devote to compressor and new features proposals always end up at the bottom of the list.

I don't hate the idea, but I prefer not having it in compressor itself, which is already way too complicated with too many options. Inheriting from `CompressorNode` and implementing it in yourself in your project seems like an approach that would work for you.

PR #596, opened 2015-02-27
fix for using block.super when there is no block.super parent

Chris7:
For whatever reason, some places define {{ block.super }} in templates that aren't extending any other templates (the root). Unfortunately, a few apps I've depended on have done this and cause offline compression to fail. This is a work-around for it. Given there is no block.super to render, skipping the extension of the current node with the block.super is not needed and I don't believe this will break anything.


diox:
Please add a test. Also, this needs to be rebased if you want it to be merged, it does not apply cleanly at the moment.

diox:
Closing per lack of answer. Please re-open if you want to resume work on this.

Issue #541, opened 2014-07-21
CSSMinFilter should not remove spaces from inside calc() expression


derek73:
CSSMinFilter removes the space from inside of `calc()` expressions, making them invalid.

`calc(100% + 20px)` becomes `calc(100%+20px)`, which is invalid. The spaces are required.

https://developer.mozilla.org/en-US/docs/Web/CSS/calc

"Note: The + and - operators must always be surrounded by whitespace"

heres:
+1

kayluhb:
+1

brmc:
Here ya go.  I hope this helps
@derek73 @heres @kayluhb

chriddyp:
thanks for the fix @brmc ! is this fix going to get merged into the official version and version bumped on pip?

diox:
2.0 has the fix (not this one in particular, instead we changed the dependencies).

PR #600, opened 2015-03-09
Bugfix: Issue #541 https://github.com/django-compressor/django-compressor/issues/541

brmc:
A new function in filters.cssmin.remove_unnecessary_whitespace, preservecalc(),  was added to preserve any calc rules from being broken during compression.
A similar solution to pseudoclasscolon() was used to maintain consistent coding style.

The tests were also updated.


brmc:
sorry about that

PR #610, opened 2015-04-14
Fix avoid getting UnicodeDecodeError

pierrei:
I ran in to this when the output of uglify-js (the compressor we're using) is outputting something that can't be read as a regular Python string. This PR fixes that bug.

When I was debugging:

```
(Pdb) "An error occured during rendering %s: %s" % (template.template_name, e.message)
*** UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 124: ordinal not in range(128)
(Pdb) "An error occured during rendering {template}: {error}".format(template=template.template_name, error=e)
'An error occured during rendering /Users/mrpi/Workspace/.../assets.html: util.error: Use console.error instead\nParse error at -:29,7\nutil.error: Use console.error instead\nUnexpected token operator \xc2\xab!\xc2\xbb, expected punc \xc2\xab(\xc2\xbb\nutil.error: Use console.error instead\nError\n    at new JS_Parse_Error (/usr/local/lib/node_modules/uglify-js/lib/parse.js:196:18)\n    at js_error (/usr/local/lib/node_modules/uglify-js/lib/parse.js:204:11)\n    at croak (/usr/local/lib/node_modules/uglify-js/lib/parse.js:679:41)\n    at token_error (/usr/local/lib/node_modules/uglify-js/lib/parse.js:683:9)\n    at expect_token (/usr/local/lib/node_modules/uglify-js/lib/parse.js:696:9)\n    at expect (/usr/local/lib/node_modules/uglify-js/lib/parse.js:699:36)\n    at parenthesised (/usr/local/lib/node_modules/uglify-js/lib/parse.js:713:9)\n    at if_ (/usr/local/lib/node_modules/uglify-js/lib/parse.js:985:20)\n    at /usr/local/lib/node_modules/uglify-js/lib/parse.js:813:24\n    at /usr/local/lib/node_modules/uglify-js/lib/parse.js:722:24\n'
```

### DIFF
--- a/compressor/management/commands/compress.py
+++ b/compressor/management/commands/compress.py
@@ -234,8 +234,7 @@ class Command(NoArgsCommand):
                 try:
                     result = parser.render_node(template, context, node)
                 except Exception as e:
-                    raise CommandError("An error occured during rendering %s: "
-                                       "%s" % (template.template_name, e))
+                    raise CommandError("An error occured during rendering {template}: {error}".format(template=template.template_name, error=e))
                 offline_manifest[key] = result
                 context.pop()
                 results.append(result)


pierrei:
I'm not sure what the formatting rules are for this repo, so I just put all in one long line. If this is not okay, please tell me and I'll update accordingly.

diox:
Please add a test that would fail without your changes, thanks.

rfleschenberg:
pierrei, can you please describe the steps to reproduce this?

rfleschenberg:
The problem arises because the exception thrown by the parser's render_node method contains its message as a bytestring, while template.template_name is a unicode object. The % operator tries to return a unicode object in that case, so it needs to decode the exception message, which fails if the message contains characters outside of the ASCII range. str.format() works differently, it will always return a bytestring (on Python 2).

Maybe the right way to fix this would actually be to to modify the code that throws the exception and have it pass the message as a unicode object.

karyon:
superseded by #804. thanks @pierrei for the PR and @rfleschenberg for helping me understand what the problem was :)

PR #614, opened 2015-04-29
Fix list formatting in docs/contributing.txt

jieter:
Fix formatting in list in docs/contributing.txt.

No list markup was generated at all.


diox:
Ah, thanks.

PR #622, opened 2015-06-20
contrib.sekizai: handle "kind" prefix with names, e.g. "js_extra"

blueyed:
It might be useful to have multiple (compressed) blocks with
django-sekizai, e.g. for global and extra/per-page JavaScript.

While you can usually just use a regular `{% compress js %}` tag for the
global JS then, there are use cases where multiple different
django-sekizai tags are useful.

This patch handles this by looking at the namespace used with the
`render_block` tag and derives the `kind` argument for `CompressorNode`
from its prefix.

This would also address #500 automatically.  /cc @nagyv

### DIFF
--- a/compressor/contrib/sekizai.py
+++ b/compressor/contrib/sekizai.py
@@ -15,4 +15,10 @@ def compress(context, data, name):
     Name is either 'js' or 'css' (the sekizai namespace)
     Basically passes the string through the {% compress 'js' %} template tag
     """
+    # Handle multiple namespaces, like "js_page" or "css_extra".
+    if name not in ('js', 'css'):
+        if name.startswith('js'):
+            name = 'js'
+        elif name.startswith('css'):
+            name = 'css'
     return CompressorNode(nodelist=Template(data).nodelist, kind=name, mode='file').render(context=context)


karyon:
although it's a little late, this would need a test, otherwise the feature is sure to decay given enough time...

karyon:
although it's a little late, this would need a test, otherwise the feature is sure to decay given enough time... and actually, some documentation as well :)

mammuth:
This is really useful! :kissing_heart: 

Wouldn't it make sense to  this more flexible. It should also work with eg. _extra_js_ and not only with _js_extra_ (like your current implementation does).

blueyed:
@mammuth 
Feel free to pick it up.. :)
(I won't have time currently to work on it)
We could have `.startswith('js_') or .endswith(`_js`)`, yes.

karyon:
could any of you confirm that https://github.com/django-compressor/django-compressor/pull/682 enables the same use cases? if yes, i'd prefer that one, since i think the string magic in this PR here is not the right approach :)

blueyed:
@karyon 
Yes, it enables the same use case, although being more verbose.

karyon:
okay, thanks. i'll close this in favor of #682.

PR #623, opened 2015-06-20
contrib.sekizai: fix double-rendering: use TextNode, not Template

blueyed:
Use `TextNode(data)` instead of `Template(data).nodelist` to parse the
template only once.

The report and fix was initially provided by Aaron McMillin.

Fixes https://github.com/django-compressor/django-compressor/issues/361
Closes https://github.com/django-compressor/django-compressor/pull/362

I've taken the fix from the initial PR (#361) and have only taken the single test that triggered the bug into the existing tests for template tags.
Not sure about how to handle the credit for this - should I amend the commit (with @aarcro's authorship?)  - but then this could be done when merging / cherry-picking it, too.

### DIFF
--- a/compressor/contrib/sekizai.py
+++ b/compressor/contrib/sekizai.py
@@ -6,7 +6,7 @@
  and: https://github.com/ojii/django-sekizai.git@0.6 or later
 """
 from compressor.templatetags.compress import CompressorNode
-from django.template.base import Template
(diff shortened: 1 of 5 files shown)


diox:
@blueyed : could you just make 2 commits in this PR one with @aarcro's authorship, and one with your as the author, with your changes ?

Issue #626, opened 2015-06-29
Compressor breaks when staticfiles added with AppConfig


jarshwah:
https://github.com/django-compressor/django-compressor/blob/develop/compressor/utils/staticfiles.py#L7 ensures that:

```
if "django.contrib.staticfiles" in settings.INSTALLED_APPS:
```

But if you've added staticfiles like so:

```
INSTALLED_APPS = (
    ...
    'django.contrib.staticfiles.apps.StaticFilesConfig',
    ...
)
```

The check fails, and compressor breaks.

rfleschenberg:
Why do you add it like that? Is that officially supported by Django?

jarshwah:
Yes - that's the new way to add applications from 1.7, using AppConfigs. If you don't specify one yourself, you get the default. But that might not always be the case, so it's better to be forwards compatible.

rfleschenberg:
Alright, thanks! I wrote a failing test and am working on a fix for this.

diox:
Should be fixed in the `develop` branch thanks to  @rfleschenberg 's patch.

PR #628, opened 2015-06-30
Use app registry to check for staticfiles on Django >= 1.7.

rfleschenberg:
This fixes issue #626.


diox:
Thanks, that looks great.

PR #632, opened 2015-07-09
Added doc to explain using compressor with facebook react

ramast:


carltongibson:
Hi @ramast. 

This looks good. It ties in with the React documentation: [Use React and JSX in Python Applications | React](https://facebook.github.io/react/blog/2013/08/19/use-react-and-jsx-in-python-applications.html)

I'm wondering if it isn't a bit heavy-weight though. Why aren't we just telling people to do something like this:

    COMPRESS_PRECOMPILERS = (
            ('text/jsx', 'cat {infile} | jsx > {outfile}'),
    )

? 

(Wondering for my own benefit.)

ramast:
Of course that is a good and simple solution and should be included in 
the doc but here is why it may not be the best solution for some users:

According to the documentation
https://facebook.github.io/react/docs/tooling-integration.html

jsx command can only be installed through npm (nodejs package manager) 
which makes npm and nodejs system wide dependencies (not just virtual 
env dependency).
If user hosting his django on a shared server that doesn't have node.js 
/ npm the solution above won't work. Even if he has these he has to know 
how to install npm packages without root access.

I suggest including it as an another option/solution

Cheers,
Ramast

carltongibson:
> which makes npm and nodejs system wide dependencies (not just virtual
env dependency).

But PyExecJS requires the JavaScript runtime to be on the system anyway — you don't get rid of that. (There's no single `pip install ...` that we can recommend to users that gets around them needed to have node, or ..., available right?)

ramast:
True, A JavaScript runtime is required.
Windows and Mac users have it pre-installed.
Linux users have 5 options to choose from (with PyV8 being the 
recommended one by PyReact)
I don't see a reason to force users to install node.js if they already 
have another JS Engine or want to go with PyV8

I want to stress that the solution you suggest is good but not for everyone.

carltongibson:
OK — that makes sense — it's good to show this integration. 

I'm imagining the 80% case to be someone with t2.micro instance running Ubuntu for whom the `apt-get` + `npm` route is the "QuickStart" they'd be after. I'm also imagining a lot of people are relying on `npm` somewhere else in their project too — I grant you not all, but a lot. Does this sound reasonable to you? 

If so, can I suggest adding the `jsx` suggestion before/as well as the full-blown PyReact approach?

ramast:
Putting the simple solution first make perfect sense to me as well.
I hate taking credit for something that I didn't come up with so would you please update the document as you suggested and push it? I already gave you write access to my fork

Cheers

carltongibson:
OK. I added the section. 

@diox What thinks you? IMO: Automatically applying precompiles like this is a great plus of compressor and React is a good example to callout. 

Is this something you think we can add? Are there changes you'd like to see? 

Thanks @ramast — good one!

ramast:
Looks good to me, Thanks carltongibson

carltongibson:
Argh! Grrr. The travis build failed, on some environments:


```
pkg_resources.DistributionNotFound: The 'setuptools' distribution was not found and is required by the application
make: *** [runtests] Error 1
ERROR: InvocationError: '/usr/bin/make test'
___________________________________ summary ____________________________________
ERROR:   py32-1.8.X: commands failed
The command "tox" exited with 1.
Done. Your build exited with 1.
```

At first glance I'm not sure what to say about that.

carltongibson:
There are a couple of different errors on the build. But none which are explained by the difference between 6b7b6aa (which passed) and 8ae9c0f43b707e99a1011329b9ea5d4918066ab7 (last commit which failed).

carltongibson:
Hey @diox — I can't explain the errors on the Travis build. 

It looks like a corrupted build environment or such: the errors are of two kinds, missing dependencies and systax errors in library code. 

Can I ask you to restart the build, possibly clearing any cached environment, if that's possible? 

Thanks!

diox:
Build restarted.

diox:
And failures still, which I believe are due to us not pinning versions of packages in `requirements.txt`. That would deserve a separate pull request, IMHO.

carltongibson:
OK. It makes sense now: errors are for Python 2.6 and 3.2 only. The 2.6 errors are due to a dictionary comprehension. I need to look up what's going on with 3.2 and the missing setuptools/mock dependencies.

I opened a new issue for this #633

diox:
This is now fixed so if you rebase this against `develop` the errors should go away.

carltongibson:
And sure enough... Success

diox:
That looks good. Have you tested both solutions yourself ? I haven't :)

carltongibson:
I’m using the `cat …` approach with perfect results.

carltongibson:
@ramast I assume you'll vouch for the custom filter approach?

ramast:
I also have tested the other solution and using the same code in production 


Sent from my Samsung device

-------- Original message --------

diox:
Cool, thanks all!

PR #635, opened 2015-07-13
Perserve spaces arround '+'

mjonker-embed:
This breaks the `calc(100% + 10px)`


diox:
Hi,

Please take that issue to rcssmin upstream: https://github.com/ndparker/rcssmin/issues/1

Once a new version of rcssmin is released with the fix we'll take it.

PR #640, opened 2015-07-29
Make a copy of offline context before using

stefanw:
The offline context dictionary can be changed by the extends node (or a subclass like django-overextends). To prevent carrying over state from previous parsings, make a copy of the offline context.

This bit me with [`django-overextends`](https://github.com/stephenmcd/django-overextends/blob/master/overextends/templatetags/overextends_tags.py#L105) which keeps state in the context between single template renderings but breaks when state is kept between many template renderings.


diox:
Thanks for the PR. Can you come up with a test for that ?

gasman:
Have added a test for this in #670.

diox:
Closing in favor of #670 then

PR #499, opened 2014-03-30
Don't continue when raising a `UnicodeDecodeError`

mattrobenolt:
Don't continue executing more code, by adding a `continue` to start the loop over. :)

PR #643, opened 2015-07-31
Don't continue when raising a `UnicodeDecodeError`

mattrobenolt:
See: #499

/cc @diox


diox:
Shouldn't the title of the commit be "Continue after catching a `UnicodeDecodeError`" instead ?

karyon:
replaced by #760 with a new commit message

PR #644, opened 2015-08-01
rCSSmin/rJSmin arg improvements

scop:
Pass keep_bang_comments=True to rCSSmin as well, and do it more cleanly with kwargs to rJSmin.


diox:
Hi, this seems like a good idea, do you think you could add a quick simple test for both ?

scop:
Sure, here goes. While at it, added bang comment to the existing cssmin test case as well.

diox:
Awesome, thanks!

PR #652, opened 2015-09-11
Refactorings

karyon:
see commit messages. also includes #643 with changed commit message because why not.

closes #519

### DIFF
--- a/compressor/base.py
+++ b/compressor/base.py
@@ -36,15 +36,17 @@ class Compressor(object):
     Base compressor object to be subclassed for content type
     depending implementations details.
     """
-    type = None
 
-    def __init__(self, content=None, output_prefix=None, context=None, *args, **kwargs):
+    def __init__(self, content=None, output_prefix=None,
+                 context=None, filters=None, *args, **kwargs):
         self.content = content or ""  # rendered contents of {% compress %} tag
         self.output_prefix = output_prefix or "compressed"
         self.output_dir = settings.COMPRESS_OUTPUT_DIR.strip('/')
         self.charset = settings.DEFAULT_CHARSET
         self.split_content = []
         self.context = context or {}
+        self.type = output_prefix or ""
+        self.filters = filters or []
         self.extra_context = {}
         self.precompiler_mimetypes = dict(settings.COMPRESS_PRECOMPILERS)
         self.finders = staticfiles.finders
--- a/compressor/base.py
+++ b/compressor/base.py
@@ -46,7 +46,7 @@ class Compressor(object):
         self.split_content = []
         self.context = context or {}
         self.extra_context = {}
-        self.all_mimetypes = dict(settings.COMPRESS_PRECOMPILERS)
+        self.precompiler_mimetypes = dict(settings.COMPRESS_PRECOMPILERS)
         self.finders = staticfiles.finders
         self._storage = None
 
@@ -205,7 +205,7 @@ class Compressor(object):
                 options = dict(options, filename=value)
                 value = self.get_filecontent(value, charset)
 
-            if self.all_mimetypes:
+            if self.precompiler_mimetypes:
                 precompiled, value = self.precompile(value, **options)
 
             if enabled:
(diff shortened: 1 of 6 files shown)


karyon:
@diox, the "remove silent exception catching" commit revealed randomly failing test. could you have a look at that? i'm not sure where to start, i can't repro this locally.

here's another test run of the same commit: https://travis-ci.org/django-compressor/django-compressor/builds/80111522

diox:
I've restarted the builds that errored but it's failing again. I have no idea why this test fails, I've never seen it fail before...

diox:
I'd prefer it this PR was splitted into multiple smaller ones, with tests covering the changes. Otherwise I have no idea what I'm merging or why.

diox:
Note: if you rebase, you should benefit from #656 which makes the tests run on travis sudo-less architecture, which means faster builds, which means faster failures to help debugging this :)

karyon:
the failures appear to be random. i've restarted the tests about 4 times, and it's always different runs that are failing, but always the same test and error.

you haven't seen this failure before because there was a "except Exception" which hid the problem. i removed the except in a commit.

opening a PR for every single commit seemed a bit overkill to me. but if you really want, i can do that later.

diox:
I see. Can you just isolate "remove silent exception catching" in a separate PR ? It'll make it easier to merge the rest first, and then we can concentrate on fixing this random failure.

karyon:
alright, i completely separated the commits now. please let me know whether you want anything changed.

compressor/base.py
```diff
@@ -36,17 +36,18 @@ class Compressor(object):
     Base compressor object to be subclassed for content type
     depending implementations details.
     """
-    type = None
 
-    def __init__(self, content=None, output_prefix=None, context=None, *args, **kwargs):
+    def __init__(self, content=None, output_prefix=None, context=None, filters=None):
```
diox: I prefer leaving `*args` and (more importantly) `**kwargs`, removing them is a backwards-incompatible change for people inheriting from it.

karyon:
done.

diox:
I'm not a fan of https://github.com/karyon/django-compressor/commit/24551f5dbd28084077eb17b9291b6140fb21de03. It removes methods that are useful for users inheriting from compressor. What was the motivation behind this change ?

karyon:
i removed render_cached because there was only one caller and it removing it reduced code complexity noticably. render_output just seemed superfluous so i inlined it, debug_mode was called nowhere so i simply removed it.

i can absolutely restore them if you want, although for render_cached i'd need to check again whether i can keep some of the simplifications, and for debug_mode i'd like to hear when it is intended to be called :)

diox:
I kinda like having `render_cached` separated so that it's easy to implement custom behaviour by inheriting from the class (the `is_offline_compression_enabled` exists for a similar purpose, I've used in the the past to make custom compress template tag nodes that are not offline cached, despite behaving normally otherwise).

The rest, and in particular the extra `debug_mode` method, can go.

karyon:
i split up the respective commit and added a deprecation warning fix. please review.

basically, the refactoring moves the decision whether to render (cached or offline) from the methods into the main method, which makes things a bit easier to read (for me at least)

karyon:
i think https://github.com/karyon/django-compressor/commit/68981ebfbf8af2a47a24a85fcee0d660a556d529#diff-ac8de32c7b59ab69b2ac7137529224f0R85 also kinda includes https://github.com/django-compressor/django-compressor/pull/519.

diox, could you confirm that the offline_manifest[key] cannot return None when key in offline_manifest?

also, since these "refactorings" apparently do fix something, do you think a test is necessary?

diox:
`offline_manifest[key]` should not return `None`. It's a simple dict (stored and loaded using a JSON file).

Since you're slightly changing behaviour (for the better), consider adding a test, yes.

karyon:
test added.

diox:
Excellent, thanks.

PR #653, opened 2015-09-11
apply CssAbsoluteFilter to precompiled css even when compression is disabled

### LINKED ISSUE #226: COMPRESS_PRECOMPILERS break css url()
Since the precompiled files are served from another location, relative links in css files do not work anymore. 

The Problem can be solved by setting the output dir to an empty string.

    COMPRESS_OUTPUT_DIR = ''

However, it would be great to have processed urls in the css file that link to the right locations of images.

### LINKED ISSUE #296: problem with relative path in less file
I'm using django-compressor with lessc in order to compile twitter-bootstrap less framework.

The issue is the following:

The compressor is not converting the relative path pointing to an image file to it's absolute path... the result is that the django server is trying to get the file from the CACHE folder:

[01/Aug/2012 21:12:16] "GET /static/CACHE/img/glyphicons-halflings.png HTTP/1.1" 404 1706

If I change manually the relative path:

@iconSpritePath:          "../img/glyphicons-halflings.png";

to:

@iconSpritePath:          "/static/bootstrap/img/glyphicons-halflings.png";

The image is loaded, but this is not a DRY solution.

I've reproduced this with 1.1.2 and 1.2a2

Cheers,
David

karyon:
this is what i was imagining. ref #331.

fixes #226, fixes #296

### DIFF
--- a/compressor/base.py
+++ b/compressor/base.py
@@ -21,6 +21,7 @@ from compressor.conf import settings
 from compressor.exceptions import (CompressorError, UncompressableFileError,
         FilterDoesNotExist)
 from compressor.filters import CachedCompilerFilter
+from compressor.filters.css_default import CssAbsoluteFilter
 from compressor.storage import compressor_file_storage
 from compressor.signals import post_compress
 from compressor.utils import get_class, get_mod_func, staticfiles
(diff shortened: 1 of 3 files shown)


karyon:
so i've tested this quite thoroughly now while developing my project and it seems to work well. @diox, what's your opinion on this approach? what's missing to get this merged?

compressor/base.py
```diff
@@ -209,20 +210,21 @@ def hunks(self, forced=False):
                 precompiled, value = self.precompile(value, **options)
 
             if enabled:
-                yield self.filter(value, **options)
+                yield self.filter(value, self.cached_filters, **options)
+            elif precompiled:
+                if CssAbsoluteFilter in self.cached_filters:
```
diox: Add a small comment explaining that/why we want to apply `CssAbsoluteFilter` even if we're disabled

diox:
Yeah, this looks good. Needs a couple tests to make sure a) we're calling `CssAbsoluteFilter` if present even when we are "disabled" b) we're not calling it if it wasn't in the list of filters to consider in the first place.

karyon:
the test is not exactly beautiful, but as i said, copypasting it is the greater sin imo :)

anything else?

diox:
Looks good. As discussed on IRC I would not have written the tests like that myself but it's fine if you prefer it that way.

PR #656, opened 2015-09-13
Remove sudo from .travis.yml to run on travis new architecture

diox:


.travis.yml
```diff
@@ -1,7 +1,5 @@
 language: python
-before_install:
-  - sudo apt-get update
-  - sudo apt-get install csstidy libxml2-dev libxslt-dev
```
diox: It might not work, we'll see what needs to be fixed.

PR #658, opened 2015-09-15
Use static dir first

karyon:
rebased #505, squashed the two commits, fixed the tests and added a comment.

as i did not understand the code completely, the comment might be a bit flaky. let me know if i should replace it.

### DIFF
--- a/compressor/base.py
+++ b/compressor/base.py
@@ -117,18 +117,19 @@ class Compressor(object):
         get_filename('css/one.css') -> '/full/path/to/static/css/one.css'
         """
         filename = None
-        # first try finding the file in the root
-        try:
-            # call path first so remote storages don't make it to exists,
-            # which would cause network I/O
-            filename = self.storage.path(basename)
-            if not self.storage.exists(basename):
-                filename = None
-        except NotImplementedError:
-            # remote storages don't implement path, access the file locally
-            if compressor_file_storage.exists(basename):
-                filename = compressor_file_storage.path(basename)
-        # secondly try to find it with staticfiles (in debug mode)
+        if not settings.DEBUG:
+            # first try finding the file in the root
+            try:
+                # call path first so remote storages don't make it to exists,
+                # which would cause network I/O
+                filename = self.storage.path(basename)
+                if not self.storage.exists(basename):
+                    filename = None
+            except NotImplementedError:
+                # remote storages don't implement path, access the file locally
+                if compressor_file_storage.exists(basename):
+                    filename = compressor_file_storage.path(basename)
+        # secondly try to find it with staticfiles (and always in debug mode)
         if not filename and self.finders:
             filename = self.finders.find(url2pathname(basename))
         if filename:
--- a/compressor/base.py
+++ b/compressor/base.py
@@ -117,8 +117,10 @@ class Compressor(object):
         get_filename('css/one.css') -> '/full/path/to/static/css/one.css'
         """
         filename = None
+        # First try finding the file using the storage class.
+        # This is skipped in DEBUG mode as files might be outdated in
+        # compressor's final destination (COMPRESS_ROOT) during development
         if not settings.DEBUG:
-            # first try finding the file in the root
             try:
                 # call path first so remote storages don't make it to exists,
                 # which would cause network I/O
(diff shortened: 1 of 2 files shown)


karyon:
do not merge yet, for some reason this gives me empty css files in my project...

compressor/base.py
```diff
@@ -117,18 +117,21 @@ def get_filename(self, basename):
...
-            if not self.storage.exists(basename):
-                filename = None
-        except NotImplementedError:
-            # remote storages don't implement path, access the file locally
-            if compressor_file_storage.exists(basename):
-                filename = compressor_file_storage.path(basename)
-        # secondly try to find it with staticfiles (in debug mode)
+        # first try finding the file in STATIC_ROOT.
+        # skip this in DEBUG mode as files might be outdated in STATIC_ROOT,
+        # and do the compression based solely on the app's static dirs instead.
```
diox: Replace the 3 lines comment with :
 `First try finding the file using the storage class. This is skipped in DEBUG mode as files might be outdated in compressor's final destination (COMPRESS_ROOT) during development`

compressor/base.py
```diff
@@ -117,18 +117,21 @@ def get_filename(self, basename):
...
+                # call path first so remote storages don't make it to exists,
+                # which would cause network I/O
+                filename = self.storage.path(basename)
+                if not self.storage.exists(basename):
+                    filename = None
+            except NotImplementedError:
+                # remote storages don't implement path, access the file locally
+                if compressor_file_storage.exists(basename):
+                    filename = compressor_file_storage.path(basename)
+        # secondly try to find it with staticfiles (and always in debug mode)
```
diox: You can remove the  `(and always in debug mode)`

compressor/tests/test_base.py
```diff
@@ -332,6 +335,52 @@ def test_get_cachekey_leading_dot(self):
...
+
+    def tearDown(self):
+        rmtree(self.tmpdir)
+        settings.STATIC_ROOT = settings.COMPRESS_ROOT = self.old_compress_root
+        delattr(settings, "STATICFILES_DIRS")
+
+    def test_filename_in_debug_mode(self):
+        # In debug mode, files should always be compressed based on the app's
+        # static directory, not the global static directory, where files can
+        # be outdated
```
diox: Replace `files should always be compressed based on the app's static directory` by `compressor should look for files using staticfiles finders only`

karyon:
done. the empty css was entirely my fault in my project, this PR works like intended for me.

compressor/tests/test_base.py
```diff
@@ -332,6 +335,52 @@ def test_get_cachekey_leading_dot(self):
...
+        self.tmpdir = mkdtemp()
+        copytree(settings.STATIC_ROOT, os.path.join(self.tmpdir, "static"))
+        self.old_compress_root = settings.COMPRESS_ROOT
+        settings.STATIC_ROOT = settings.COMPRESS_ROOT = os.path.join(self.tmpdir, "static")
+        settings.STATICFILES_DIRS = [self.old_compress_root]
+
+    def tearDown(self):
+        rmtree(self.tmpdir)
+        settings.STATIC_ROOT = settings.COMPRESS_ROOT = self.old_compress_root
+        delattr(settings, "STATICFILES_DIRS")
```
diox: weird overriding of settings here and in `setUp`

karyon: you mean this old_compress_root thing? that's done all over the place in django-compressor, and i don't know a better way to do it when the new setting value is not known before (so you can't use the decorator.

or do you mean the delattr thing? i don't know whether the original PR did that, it should be possible to save the original value with an old_* variable and restore it here, just like for the other settings

diox: I meant both, I forgot we still had these ugly setting changes... We could use the context manager... but let's not worry about that now.

Can you replace the `delattr` with the same mechanism as you suggest though, copying to `old_*` variable ? That way it's still ugly but somewhat coherent :)

karyon: it actually threw an attribute error when i tried that, so there's why the original PR did that :)

instead i had another look at how the context manager was used in the OfflineTestCaseMixin, finally understood it and tried that. seems to work well.

karyon:
oh wait i think i screwed up the rebase

karyon:
fixed. the new TestCase ended up in the middle of another TestCase and kinda took over the test at the bottom from there

PR #662, opened 2015-09-21
Update reactjs.txt to use babel instead of jsx

alecklandgraf:
React has adopted `babel` over their `jsx` for compiling *jsx* files to *js*: http://facebook.github.io/react/docs/tooling-integration.html#productionizing-precompiled-jsx


diox:
thanks!

PR #667, opened 2015-09-25
Refactor offline tests to make them always use override_settings

diox:
Also fixed flake8 issues (preparing for the day when we remove the various ignored errors from the Makefile)


compressor/tests/test_offline.py
```diff
@@ -30,7 +30,7 @@
 # The Jinja2 tests fail on Python 3.2 due to the following:
 # The line in compressor/management/commands/compress.py:
 #     compressor_nodes.setdefault(template, []).extend(nodes)
-# causes the error "unhashable type: 'Template'"
+# causes the error 'unhashable type: 'Template''
```
diox: Yeah, I also changed all the quotes to single quotes, because that's my own preference :-)

PR #683, opened 2015-10-22
Ignoring django.template.base.TemplateDoesNotExist of get_template_sources calls

Alcolo47:
Useful for custom template loader


karyon:
could you explain the use case for this? also, the branch needs a rebase and a test :)

karyon:
no reply, closing.

PR #684, opened 2015-10-23
Change to suggest django-storages-redux in docs

aericson:
The original django-storages is a dead project that haven't seen any
updates since early 2014, including a lot of unmerged pull requests.
Django-storages-redux is a fork well accepted, under active
development and python 3 support. We should link to it instead of the
old one.

More information on the 'Why Fork?' section on the README of
https://github.com/jschneier/django-storages


diox:
Yeah, I agree. Thanks for doing this.

PR #685, opened 2015-10-23
Fix various deprecation warnings on Django 1.7/1.8

spookylukey:
Individual commits messages have more info.

There are a few still warnings unaddressed, because they require more work:

* New 'TEMPLATES' setting: https://docs.djangoproject.com/en/1.8/ref/templates/upgrading/#upgrading-templates-to-django-1-8

  If you remove 'TEMPLATE_*' from compressor/test_settings.py (for Django 1.8), as suggested by the instructions above, you get errors because 'TEMPLATE_DIRS' and 'TEMPLATE_LOADERS' are used explicitly by django_compressor.

* Removal or NoArgsCommand - https://docs.djangoproject.com/en/1.8/releases/1.8/#django-core-management-noargscommand

  It is not obvious to me how to succinctly fix this in a way that works the same as it did before but doesn't generate a warning on Django 1.8.

### DIFF
--- a/compressor/base.py
+++ b/compressor/base.py
@@ -2,9 +2,8 @@ from __future__ import with_statement, unicode_literals
 import os
 import codecs
 
+import django
 from django.core.files.base import ContentFile
-from django.template import Context
-from django.template.loader import render_to_string
 try:
     from importlib import import_module
 except:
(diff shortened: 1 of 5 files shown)


diox:
That's pretty great, thanks. We've been discussing about deprecation warnings in a couple other issues, we may end up finding a solution for those you didn't address there.

PR #692, opened 2015-11-03
Test with Python 3.5

scop:


nikolas:
It looks like django 1.9 is failing because of the `add_to_builtins` deprecation. You can do something like this:

```py
def add_to_builtins_compat(name):
    if django.VERSION < (1, 9):
        from django.template.base import add_to_builtins
        add_to_builtins(name)
    else:
        engines['django'].engine.builtins.append(name)
```

scop:
Yes, well, to me the point of this PR was to just add the tests, fixing possible failures could be done in another.

BTW I'm not sure the way you proposed to do it is an officially sanctioned one, see https://docs.djangoproject.com/en/1.9/releases/1.9/#django-template-base-add-to-builtins-is-removed

Oh, and #697 contains some overlapping code with this PR, so I likely won't touch this further until I know which way the maintainers want to go with this.

diox:
Hey, sorry, I had not seen this PR. Can you just include the Python 3.5 stuff ? We'll take care of Django 1.9 in #697.

scop:
Ok, I removed the Django 1.9 things, leaving only the Python 3.5 stuff and rebased against current develop.

diox:
Thanks!

PR #695, opened 2015-11-17
Convert paths to strings before passing them to os.walk

tino:
Otherwise os.walk will raise TypeError: coercing to Unicode: need string or buffer, Origin found with the django 1.9 release candidate. Django now passes Origin classes (https://docs.djangoproject.com/en/1.9/ref/templates/api/#django.template.base.Origin). Using Origin.name would probably be better but I don't know if that works in earlier versions.


outime:
:+1:

PR #700, opened 2015-12-04
Deal with dependencies

### LINKED ISSUE #541: CSSMinFilter should not remove spaces from inside calc() expression

### LINKED ISSUE #586: Large CSS file Online Compression crashing Django
Today I was planning on switching from Foundation to Semantic UI. Semantic CSS file is huge compared to say bootstrap or foundation and so I tried to compress it via Online Compression, but Django crashed because compression was taking too long. I've circumvented this by using offline compression which still takes around 1-2 mins but it eventually works. I was wondering why this was happening when doing it during online compression. Also, is there a way to have successful online compression for large files without crashing django? I wasn't sure if this was a bug in the django-compressor itself or was because it was trying to compress a large file in the request/response cycle.

My compressor related settings:

	# Django Compressor Settings
	COMPRESS_ENABLED = True

	COMPRESS_OUTPUT_DIR = ''

	COMPRESS_CSS_FILTERS = [
	     'compressor.filters.cssmin.CSSMinFilter',
	     'compressor.filters.css_default.CssAbsoluteFilter',
	]
	COMPRESS_JS_FILTERS = [
	     'compressor.filters.jsmin.JSMinFilter'
	]

	COMPRESS_OFFLINE = True

### LINKED ISSUE #664: deal with the dependencies

karyon:
fixes #664. as discussed, without any deprecation warnings.

closes #541, closes #600, closes #655, and closes #586 as the cssminfilter is now removed.


Makefile
```diff
@@ -10,7 +10,7 @@ runtests:
 	coverage run --branch --source=compressor `which django-admin.py` test --settings=compressor.test_settings compressor
 
 coveragereport:
-	coverage report --omit=compressor/test*,compressor/filters/jsmin/rjsmin*,compressor/filters/cssmin/cssmin*,compressor/utils/stringformat*
```
diox: yay

diox:
A few flake8 errors are causing the build to fail:
compressor/parser/beautifulsoup.py:16:84: F821 undefined name 'err'
compressor/tests/test_filters.py:6:1: F401 'textwrap' imported but unused
compressor/tests/test_filters.py:10:1: F401 'unittest' imported but unused
compressor/tests/test_filters.py:16:1: F401 'find_command' imported but unused

karyon:
fixed.

diox:
Awesome!

PR #702, opened 2015-12-11
Implement handle() method on compress command in place of handle_noargs().

rvause:
`BaseCommand` requires a `handle()` method to be implemented.


diox:
Whoops! Thanks.

karyon:
my bad! could you apply the same change to mtime_cache?

karyon:
dang, seconds too late :)

diox:
I'm going to fix mtime_cache and add tests

PR #705, opened 2015-12-15
Use django.utils.six instead of six

scop:


karyon:
do you get an import error? i wonder why it worked for me, because i cannot remember installing six...

scop:
Nope, no import error, just happened to notice otherwise.

karyon:
ah, it probably works because some dependency depends on six.

diox:
Thanks!

PR #710, opened 2015-12-17
django 1.9 fix template loader Origin

dmorina:
a tiny change to fix the os.walk `path` parameter in Django 1.9, the problem occurs because the template loaders are yielding an Origin instance and not a string(template name), therefore we need the `name` property

see:
https://github.com/django/django/blob/stable/1.9.x/django/template/loaders/filesystem.py#L47


dmorina:
nevermind just realized it was included in #701 update functions
closing

PR #712, opened 2015-12-23
Bump overextends in tests to 0.4.1, which supports Django 1.9

diox:


diox:
Looks like with the new test settings I'm now able to reproduce #617, this will need to be fixed. There are other errors though, not sure if related yet.

diox:
Closing, let's just remove overextends, it was a good idea to have these tests but they are causing more trouble than they are worth.

PR #713, opened 2015-12-23
jinja2 backend does not have 'engine' attribute. Work-around to get compresso…

nwilson5:
jinja2 backend does not have 'engine' attribute. Work-around to get compressor working for jinja2 templates in django 1.8. Update JINJA2_GET_ENVIRONMENT() to retrieve correct environment for jinja2 backend.

### DIFF
--- a/compressor/conf.py
+++ b/compressor/conf.py
@@ -2,6 +2,7 @@ from __future__ import unicode_literals
 import os
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
+from django.template.utils import InvalidTemplateEngineError
 
 from appconf import AppConf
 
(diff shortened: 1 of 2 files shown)


diox:
Add a `# flake8: noqa` to avoid this:
`compressor/conf.py:77:13: F401 'jinja2' imported but unused`

diox:
Looks like it breaks a bunch of tests :/

nwilson5:
Not sure how the tests work, but if the tests always expect it to have django templates or something that's the only reason I could see it breaking. I could've made it only have a work-around for the jinja2 engine and the tests would likely not have broken (?), although I think that's "incorrect" since it would use a single parser for both cases. Since you can have both django templates and jinja templates you can't use one parser, so I forced all the previous logic behind an if statement in the case where it's jinja2. Otherwise that code was the same.

Anyways it works for my purposes... take it or leave it.

diox:
The tests should not expect to have always django templates or something like that. They just run compress with the engines that are supposed to be supported for each test. 

I'm not sure what's going on yet, but I think there might be an issue with the patch if compress ends up not finding the templates like it does at the moment.

diox:
Thanks for the PR, I managed to get it to work in #715.

PR #714, opened 2015-12-24
Remove test using django-overextends to simplify codebase

diox:


diox:
overextends does some non trivial things that mess up with our tests, and it's only there to add coverage for one very minor line of code, essentially.

PR #726, opened 2016-01-19
return jinja2 template environment per docs

### LINKED ISSUE #637: TemplateDoesNotExist is raised when using native Jinja2 support in Django 1.8
Django 1.8 has native Jinja2 template support built-in. Therefore, I am not using Jingo, Coffin or django-jinja. In using the native Jinja2 template support option, however, there seems to be some pretty significant underlying differences in the way that context data is handled and in how Django locates Django app templates when the native Jinja2 is enabled.

These are my Django settings for the templates:
```
...
TEMPLATES = [
    {
        'BACKEND': 'django.template.backends.jinja2.Jinja2',
        'DIRS': [
            str(DJANGO_APPS_ROOT_DIR.path('templates')),
        ],
        'APP_DIRS': True,
        'OPTIONS': {
            'environment': 'oiaccount.common.template.jinja2.conf.environment',
            'extensions': [
                'jinja2.ext.i18n',
                'jinja2.ext.autoescape',
                'compressor.contrib.jinja2ext.CompressorExtension'
            ]
        }
    },
]
...
```

Using these settings and attempting to use django-compressor will raise the following exception when attempting to render a template. For example, this is the exception raised when the parser encounters a `{% compress css %}` block:

`django.template.base.TemplateDoesNotExist: compressor/css_file.html`

**Workaround:**
To fix this issue, I had to modify the django_compressor package in two (2) ways. If both these changes are not done, a `TemplateDoesNotExist` exception will be raised.

1. When the `TEMPLATES[APP_DIRS]` Django setting is True, Django looks into each Django app installed for each app's necessary templates. However, when Jinja2 is enabled, it looks for a `jinja2` top-level directory in the app. Therefore, I had to add a copy of the django-compressor template set into `/compressor/jinja2/compressor`. See the picture below.
screen shot 2015-07-20 at 1 24 13 pm
2. Unfortunately, probably due to the fact that `context_instance` argument is deprecated starting in Django 1.8 (https://docs.djangoproject.com/en/1.8/topics/http/shortcuts/#render-to-response), enabling Jinja2 seems to totally prevent the `compressor/css_file.html` from being found. To fix this, change:
```
return render_to_string(template_name, context_instance=final_context)
```
at https://github.com/django-compressor/django-compressor/blob/develop/compressor/base.py#L345
to
```
return render_to_string(template_name, context=final_context.flatten())
```

**Caveats:**
1. The additional copy needed of the template set seems pretty rough around the edges, but I have yet to figure out a better method.
2. I am not sure if this negatively impacts the situation where Django Templates (DTL) and not Jinja2 templates are used.

Any comments on my approaches?

I can make a pull request if the changes that I have made are ultimately acceptable.

thenewguy:
Use the alias 'jinja2' to look up the jinja2 template engine per https://docs.djangoproject.com/en/1.9/topics/templates/#django.template.loader.engines

fix #637

### DIFF
--- a/compressor/conf.py
+++ b/compressor/conf.py
@@ -73,13 +73,10 @@ class CompressorConf(AppConf):
 
     # Returns the Jinja2 environment to use in offline compression.
     def JINJA2_GET_ENVIRONMENT():
-        alias = 'Jinja2'
+        alias = 'jinja2'
         try:
-            from django.template.loader import _engine_list
-            engines = _engine_list(alias)
-            if engines:
-                engine = engines[0]
-                return engine.env
+            from django.template import engines
+            return engines[alias].env
         except InvalidTemplateEngineError:
             raise InvalidTemplateEngineError(
                 "Could not find config for '{}' "


compressor/conf.py
```diff
@@ -73,13 +73,10 @@ class CompressorConf(AppConf):
...
-        alias = 'Jinja2'
+        alias = 'jinja2'
         try:
-            from django.template.loader import _engine_list
-            engines = _engine_list(alias)
-            if engines:
-                engine = engines[0]
-                return engine.env
+            from django.template import engines
+            return engines[alias].env
```
diox: I'm all for that change, but it means the exception we catch below isn't quite right. It should be `IndexError`, right ? Do you think you can try adding a test for that ? Something that overrides settings to omit jinja2 template configuration and then tries to access `COMPRESS_JINJA2_GET_ENVIRONMENT` ?

thenewguy: I will check again but I thought when I looked it up before the engines object raises the custom exception instead of indexerror. I'm travelling so I am not going to be able to get to this right away

diox: You're right actually, totally forgot about that.

Issue #718, opened 2015-12-31
Recursion error when extending Django admin templates and compressing offline


chriskief:
Compressor throws a recursion error when trying to compress offline and admin extension templates are present:

```
$ python /www/360i/artifact/manage.py compress

Traceback (most recent call last):
  File "/www/360i/artifact/manage.py", line 11, in <module>
    execute_from_command_line(sys.argv)
  File "/www/360i/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 350, in execute_from_command_line
    utility.execute()
  File "/www/360i/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 342, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/www/360i/local/lib/python2.7/site-packages/django/core/management/base.py", line 348, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/www/360i/local/lib/python2.7/site-packages/django/core/management/base.py", line 399, in execute
    output = self.handle(*args, **options)
  File "/www/360i/local/lib/python2.7/site-packages/compressor/management/commands/compress.py", line 286, in handle
    self.compress(sys.stdout, **options)
  File "/www/360i/local/lib/python2.7/site-packages/compressor/management/commands/compress.py", line 181, in compress
    nodes = list(parser.walk_nodes(template))
  File "/www/360i/local/lib/python2.7/site-packages/compressor/offline/django.py", line 150, in walk_nodes
    for node in self.walk_nodes(node, original):
  File "/www/360i/local/lib/python2.7/site-packages/compressor/offline/django.py", line 146, in walk_nodes
    for node in self.get_nodelist(node, original):
  File "/www/360i/local/lib/python2.7/site-packages/compressor/offline/django.py", line 128, in get_nodelist
    return handle_extendsnode(node, block_context=None, original=original)
  File "/www/360i/local/lib/python2.7/site-packages/compressor/offline/django.py", line 41, in handle_extendsnode
    return handle_extendsnode(node, block_context, original)
  File "/www/360i/local/lib/python2.7/site-packages/compressor/offline/django.py", line 41, in handle_extendsnode
    return handle_extendsnode(node, block_context, original)
  { ... CONTINUED ... }
  File "/www/360i/local/lib/python2.7/site-packages/compressor/offline/django.py", line 30, in handle_extendsnode
    context = Context()
[...]

PR #727, opened 2016-01-19
Regression test for #718 ("extends" recursion error)

yprez:


compressor/tests/test_templates/test_extends_recursion/admin/index.html
```diff
@@ -0,0 +1 @@
+{% extends "admin/index.html" %}
```
yprez: This is what causing it to fail, commenting it out fixes the tests.

yprez:
Not the prettiest way to test this, but it was the best I could come up with at this point...

diox:
This is helpful, thanks. No idea when I/someone will have the time to fix it, but having a clear failing test is a great start.

yprez:
I tried finding the source of the bug yesterday but without much luck. I'll try taking a fresh look at it again at some point...

yprez:
Superseded by #728

PR #728, opened 2016-01-20
Fix offline extends recursion

cbjadwani:
This fixes #718 and uses tests added in #727.

Passing an empty context for each extended template was breaking django's mechanism to find correct template if the name of the parent template is same as base template by keeping a stack of 'seen' templates. This PR fixes that by using a single context when finding parent templates.

PR #731, opened 2016-01-22
Fix Travis build

### LINKED ISSUE #729: Latest version of virtualenv (14.0.0 2016-01-19) breaks the build
After upgrading virtualenv to 14.0.0 the tox (/Travis) run fails with: "no such option: --download-cache".

Possible fixes I can think of:

1. Freeze the virtualenv version in tox: `pip install "virtualenv<14"`
2. Remove the `downloadcache` from tox.ini

I tested both, but I don't like neither of them...

This is probably something new since the virtualenv came out yesterday, and I couldn't find related issues. I suppose they will pop up soon.

So option 3 is to wait until this is fixed by tox or virtualenv.

yprez:
Fixes tox / travis build issues, caused by the latest release of pip and virtualenv.

Fixes #729

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: python
 sudo: false
 install:
+  - if [[ $TOXENV == py32-1.8.X ]]; then pip install pip\<8.0.0 virtualenv\<14.0.0; fi
   - pip install tox
 script:
   - tox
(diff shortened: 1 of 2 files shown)


diox:
Obsolete now that I've merged #728.

PR #733, opened 2016-01-22
compress multiple engines at once

thenewguy:
Submitted for ci


thenewguy:
@diox Let me know what you think about this.  Working from a tablet at the airport so I don't have my full coding stack available for debugging.

diox:
Well I certainly would be unable to code from a tablet, at the airport, so kudos !

Looping over `engines` in the handle method, calling `self.compress` multiple times is not how I envisioned it originally - I thought you'd loop over the engines in the compress method itself - but I guess that works.

I think the test is too fragile however. `manifest_django` and `manifest_jinja2` are identical in your case, so I'm not sure the test is very useful.

Have a look at the `test_offline.py` tests - they set specific template paths to ensure a subset of test templates are considered for each test class, and I think we need something like this here. Ideally you want a test that would work when we have jinja and django-specific templates involved, possibly with different outputs.

thenewguy:
@diox it shouldn't be any trouble to move the logic. I went t path of least resistance but I think you're definitely right about where the logic should be.  I wasn't sure about the way the templates were discovered for testing.  Thanks for pointing me towards something to compare against.  I'll touch that up soon

thenewguy:
@diox So I've updated the test.  Please take a look.  I've addressed the testing concerns you mentioned.  The only other way to simplify the test would be to hard code the expected result but I don't think that is what you were looking for.

Also, I would like to leave the engine logic in the command's handle method at first if you're good with that.  It can be moved into the compress method in a second pull request but it will require a little bit of refactoring that I do not have time for at the moment.

thenewguy:
The tests that failed are due to the import I think.  The test I added works locally for me.  I need to move the command test I added to the other file instead of importing the base command

thenewguy:
Passing now.  I was running the test locally with the complete dotted path to the method which is why it was not showing me the errors from the inherited methods locally.  Copy/paste had me scratching my head heh

diox:
Sorry for not replying earlier. I feel like it needs more tests - your test is a step in the good direction but not enough to prove that everything works correctly IMHO: you are not proving that the files are correctly generated for all templates in all engines, and make sure the hashes are correct.

I don't have the time to elaborate more at the moment.

thenewguy:
Are you saying you want me to compare the manifest against a hardcoded expected result?  I've run into problems with compressor in the past where newlines created different hashes on Windows and Linux so I didn't think that was the way to go (but this was a long time ago).  Other than that I am not really sure what the test doesn't cover.

@diox I will explain my reasoning. When you have time, if you tell me where I am wrong then it would help and I will improve the testing.

The existing tests should already prove that the django and jinja2 template engines work as expected when used individually.  So the manifests generated by `call_command('compress', engines=["django"], **opts)` and `call_command('compress', engines=["jinja2"], **opts)` can be assumed correct.  So both `manifest_django` and `manifest_jinja2` from `get_offline_manifest()` are correct.

What is new is this `call_command('compress', engines=["django", "jinja2"], **opts)` which produces a manifest containing both the jinja2 template output and the django output automatically merged into one manifest.

Since `manifest_django` and `manifest_jinja2` are known to be correct as confirmed by the preexisting tests in compressor, we know that the manifest for `engines=["django", "jinja2"]` should be equal to the union of these two manifests.  This is created and stored as `manifest_both_expected`

The lines `self.assertNotEqual(manifest_django, manifest_jinja2)`, `self.assertTrue(manifest_django)`, and `self.assertTrue(manifest_jinja2)` ensures that each engine produced unique results (achieved by different javacript string in the jinja2 and django templates) and neither manifest was empty ensuring that the javascript was found and compressor actually did something.

Then the real test is `self.assertEqual(manifest_both, manifest_both_expected)`.  This proves that the manifest created by the union of manifests created by each engine is equal to the manifest created by running both engines together.

What am I missing?

diox:
If you look at the other offline tests you'll notice they compare against a hash, effectively comparing against an expected result. This is a quick way to verify that not just the manifest is correct, but also the produced files.

thenewguy:
@diox Now verifies manifests against expected hash output

compressor/tests/test_offline.py
```diff
@@ -655,3 +656,31 @@ def _test_offline(self, engine):
...
+        call_command('compress', engines=["django"], **opts)
+        manifest_django = get_offline_manifest()
+
+        call_command('compress', engines=["jinja2"], **opts)
+        manifest_jinja2 = get_offline_manifest()
+
+        call_command('compress', engines=["django", "jinja2"], **opts)
+        manifest_both = get_offline_manifest()
+
+        self.assertEqual(manifest_django, {'8464063aa0729700fca0452e009582af': '<script type="text/javascript" src="/static/CACHE/js/662b9ce354e4.js"></script>'})
```
diox: Please wrap that at 79 columns (we still have files not respecting that in the project, but I tried to keep this one as clean as possible)

thenewguy: Will do.  Adding one more test to verify commandline syntax

thenewguy:
@diox All tests pass. Think its good to go now

thenewguy:
@diox slightly unrelated to this pr, but would you consider adding appveyor ci to the project to test on windows?  5 tests fail for me on windows so I find myself only running the test I am working on before pushing

karyon:
Rebased in #806

PR #737, opened 2016-01-30
Add multiprocessing option in compress command

cbjadwani:
Re-implements #353 using `multiprocessing.Pool` to use fixed number of processes instead of creating a new one for each compress node.

### DIFF
--- a/compressor/management/commands/compress.py
+++ b/compressor/management/commands/compress.py
@@ -249,7 +249,7 @@ class Command(BaseCommand):
             finally:
                 pool.close()
                 pool.join()
-        results = [r for r in results if r]
+        results = [r for r in results if r is not None]
 
         write_offline_manifest(offline_manifest.copy())
 
--- a/compressor/management/commands/compress.py
+++ b/compressor/management/commands/compress.py
@@ -5,6 +5,8 @@ import sys
 from collections import OrderedDict
 from fnmatch import fnmatch
 from importlib import import_module
+import multiprocessing
+import traceback
 
 import django
 from django.core.management.base import BaseCommand, CommandError
@@ -47,6 +49,10 @@ class Command(BaseCommand):
                                  "(which defaults to STATIC_ROOT). Be aware that using this "
                                  "can lead to infinite recursion if a link points to a parent "
                                  "directory of itself.", dest='follow_links')
+        parser.add_argument('--processes', '-p', default=1, type='int',
+                            dest='processes',
+                            help='Number of parallel processes to use to compress blocks '
+                            '(default 1).'),
         parser.add_argument('--engine', default="django", action="store",
                             help="Specifies the templating engine. jinja2 or django",
                             dest="engine")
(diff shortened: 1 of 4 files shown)


compressor/management/commands/compress.py
```diff
@@ -47,6 +49,10 @@ def add_arguments(self, parser):
                                  "(which defaults to STATIC_ROOT). Be aware that using this "
                                  "can lead to infinite recursion if a link points to a parent "
                                  "directory of itself.", dest='follow_links')
+        parser.add_argument('--processes', '-p', default=1, type='int',
+                            dest='processes',
+                            help='Number of parallel processes to use to compress blocks '
+                            '(default 1).'),
```
karyon: @diox any reason to not set the default to "number of processors"?

compressor/management/commands/compress.py
```diff
@@ -208,47 +214,48 @@ def compress(self, log=None, **options):
...
-                context = Context(init_context)
                 template._log = log
                 template._log_verbosity = verbosity
+                template_tasks = ((template, parser, node, context_dict,
+                        offline_manifest, offline_manifest_lock)
+                            for node in nodes)
+                if processes > 1:
+                    tasks += template_tasks
+                else:
+                    results += (_do_render(*task) for task in template_tasks)
```
karyon: could you make a quick check whether the lock and this mapping stuff creates any noticeable slowdown for only one thread? because if not, maybe we could just remove the `processes == 1` case to reduce complexity

compressor/management/commands/compress.py
```diff
@@ -208,47 +214,48 @@ def compress(self, log=None, **options):
...
+
+        if tasks:
+            pool = multiprocessing.Pool(processes=processes)
+            try:
+                results = pool.map(_do_render_task, range(len(tasks)))
```
karyon: why not map onto the tasks directly? (disclaimer, i don't really know the multiprocessing module...)

compressor/management/commands/compress.py
```diff
@@ -208,47 +214,48 @@ def compress(self, log=None, **options):
...
+            finally:
+                pool.close()
+                pool.join()
+        results = [r for r in results if r]
```
karyon: `if r is not None`, let's not assume r cannot be `[]` or `""` (actually i don't know whether that can happen but let's be safe)

compressor/management/commands/compress.py
```diff
@@ -284,5 +291,45 @@ def handle(self, **options):
...
+
+def _do_render(template, parser, node, context_dict, manifest, manifest_lock):
+    init_context = parser.get_init_context(context_dict)
+    context = Context(dict_=init_context)
+
+    rendered = parser.render_nodelist(template, context, node)
+    key = get_offline_hexdigest(rendered)
+
+    if key in manifest:
+        return None
```
karyon: do you know whether this is a vital optimization? i.e., if it doesn't run much slower without this, we could just remove this and the lock altogether, let the tasks return tuples like `(key, result)` and build the dictionary in serial afterwards

compressor/management/commands/compress.py
cbjadwani: This indirection to get arguments of `_do_render` is necessary because arguments of the target function of Pool.map must be 'pickle'able.

compressor/management/commands/compress.py
cbjadwani: I think this avoids a lot of duplicate work in cases where compressed block is inherited in multiple templates from a base template.  There is also a test case for this `OfflineCompressSkipDuplicatesTestCase`.

compressor/management/commands/compress.py
cbjadwani: ok, I will to do some benchmarking for single process and in-process implementations.  I implemented this way as I wanted to avoid introducing multiprocessing in the default use case.

compressor/management/commands/compress.py
diox: I don't have a strong opinion one way or the other. I don't think backwards-compatibility matters here, if that's what you were worried about.

compressor/management/commands/compress.py
diox: Is there any reason why we'd want to avoid multiprocessing ?

cbjadwani: The single process special case is not necessary but AFAIU using process pool for single process would add unnecessary overhead - creating a new process and serializing/sending data between main process and pool process.

Also if there was any bug in multiprocess code this would allow a safe fallback.

(I will try to do some basic benchmarking during weekend)

karyon: ping :)

karyon:
hey @cbjadwani, please let us know whether you'll continue with this :)

karyon:
For the record, i tried to rebase this but gave up. c0010bf2ae4064c11cfee2387f6e06c3880fc6d3 changed the same parts quite fundamentally. it's probably not much work to adapt this PR, but I wasn't familiar enough with the code.

karyon:
closing, #912 is probably closer to being mergeable.

PR #740, opened 2016-02-03
Updated required django-appconf version

jonklo:
django-appconf version 0.4 is not compatible with Django 1.9. The old package is still used when upgrading a project to Django 1.9, so this requirement breaks 1.9 compatibility. 

Fixed by using the latest version of django-appconf.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -141,7 +141,7 @@ setup(
     ],
     zip_safe=False,
     install_requires=[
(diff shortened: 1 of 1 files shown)


diox:
Good catch!

diox:
Thanks, sorry for forgetting to bump that.

PR #745, opened 2016-02-27
Fixing 1.10 compatibility issues

jaddison:
Django 1.10 dropped optparse in favour of argparse as per their deprecation policy. the `compress` management command is now updated.

Additionally, `TEMPLATE_*` parameters have also been dropped with 1.10. the `TEMPLATES` setting is now used. A small change relating to the availability of loaders was modified in the `compress` command.


karyon:
thanks!

PR #749, opened 2016-03-08
Updating remote storages documentation for CachedS3BotoStorage bug.

BriggsThompson:
The example in the docs doesn't quite work as the file stored in
the "content" variable is already closed, and ends up spitting out
encoded data to the destination files. It was properly being sent
to S3, but not getting saved locally.

This saves the file locally first, then using local storage, opens
the file once again for the S3BotoStorage save call.

### DIFF
--- a/docs/remote-storages.txt
+++ b/docs/remote-storages.txt
@@ -66,8 +66,8 @@ apps can be integrated.
                 "compressor.storage.CompressorFileStorage")()
 
         def save(self, name, content):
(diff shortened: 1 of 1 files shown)


karyon:
as it's not quite obvious what the problem is, could you add a test for that?

karyon:
scratch that, i just realized this code is inside the documentation, sorry.

karyon:
thanks!

Issue #750, opened 2016-03-12
CompilerFilter fails with UnicodeDecodeError when cache used


akolpakov:
Very related to bug #311 and pull request #312. In that commit problem was fixed only for case when cache is not used or missed.
But in case when cache was hitted `UnicodeDecodeError` exception is raised:

```
...
  File "...sandbox/lib/python2.7/site-packages/compressor/templatetags/compress.py", line 127, in render
    return self.render_compressed(context, self.kind, self.mode, forced=forced)
  File "...sandbox/lib/python2.7/site-packages/compressor/templatetags/compress.py", line 103, in render_compressed
    rendered_output = compressor.output(mode, forced=forced)
  File "...sandbox/lib/python2.7/site-packages/compressor/css.py", line 49, in output
    ret.append(subnode.output(*args, **kwargs))
  File "...sandbox/lib/python2.7/site-packages/compressor/css.py", line 51, in output
    return super(CssCompressor, self).output(*args, **kwargs)
  File "...sandbox/lib/python2.7/site-packages/compressor/base.py", line 292, in output
    output = '\n'.join(self.filter_input(forced))
  File "...sandbox/lib/python2.7/site-packages/compressor/base.py", line 232, in filter_input
    for hunk in self.hunks(forced):
  File "...sandbox/lib/python2.7/site-packages/compressor/base.py", line 215, in hunks
    basename=basename)
  File "...sandbox/lib/python2.7/site-packages/compressor/base.py", line 307, in handle_output
    return output_func(mode, content, forced, basename)
  File "...sandbox/lib/python2.7/site-packages/compressor/base.py", line 319, in output_file
    self.storage.save(new_filepath, ContentFile(content.encode(self.charset)))
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 126709: ordinal not in range(128)
```
I faced with this exception while I used memcached for caching.

karyon:
fixed by #751

PR #751, opened 2016-03-12
#750 CompilerFilter fails with UnicodeDecodeError when cache is used

akolpakov:
Fix bug #750.

### DIFF
--- a/compressor/filters/base.py
+++ b/compressor/filters/base.py
@@ -221,7 +221,7 @@ class CachedCompilerFilter(CompilerFilter):
             key = self.get_cache_key()
             data = cache.get(key)
             if data is not None:
-                return data
+                return smart_text(data)
             filtered = super(CachedCompilerFilter, self).input(**kwargs)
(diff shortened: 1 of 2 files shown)


karyon:
hey, any chance you can whip up a test for that analogous to the one in #312?

akolpakov:
Done!
Test with emulation of memcached which returns string regardless that we set unicode to it.

karyon:
thanks! i'll have to bother you once more, but could you make the test provoke the actual failure (the crash)? right now (as far as i can see) it asserts that some string has some type, but that itself wouldn't be a problem if some other part didn't crash because of that. or am i misunderstanding something?

akolpakov:
Yes. This is the cause of crash. Crash log itself you can see in bug #750.
But this is unit tests, which check only particular small functionality. I don't see the reason to put such huge test which includes chain which cross-cover whole plugin.

karyon:
alright, thanks!

PR #756, opened 2016-04-15
Use proper options dict for conversion to tuple in CompilerFilter

grzegorzswica:
Fixed convertions of options from dict to tuple.

Previously kwargs were duplicated and proper options were dropped.


karyon:
how did you catch that, was there a specific error you encountered?

grzegorzswica:
@karyon `KeyError` when formatting command with options. `command = self.command.format(**options)`

karyon:
thanks! any chance you can provide a test case for that?

grzegorzswica:
Test case added.

karyon:
thanks!

Issue #743, opened 2016-02-16
Django 1.9 support


vladir:
I'm thinking into upgrade a project that uses django-compressor very soon to Django 1.9. Is this already supported? Thanks in advance.

diox:
Yes, the latest release supports Django 1.9.

vladir:
Thanks a lot @diox, good to know that.

PR #757, opened 2016-04-18
Reword README

karyon:
Triggered by https://github.com/django-compressor/django-compressor/issues/743, i wanted to put the supported django versions somewhere, and then i thought one of the coolest features (precompilers) is missing from the readme and so i added that and some smaller changes as well.


diox:
Looks good to me.

Issue #674, opened 2015-10-02
Support inheritance in offline compression using a variable


diox:
https://docs.djangoproject.com/en/1.8/ref/templates/builtins/#std:templatetag-extends

This used to work, but wasn't tested so it broke when we added multiple offline contexts in #572 and I found out while doing some tests in #673 and #670. I disabled it entirely in de3afb17ab102430e9d3c97034f6016c6a0d3037 to be in a more stable, homogeneous situation.

I'd like to fix it, but I'd also like to keep multiple contexts, so we need to refactor things a little bit, because right now we only go through the parent nodes once, we need to do that for each context instead, without mixing up the contexts when rendering the compress nodes later.

diox:
Both a bug and a proposal since it used to work :)

diox:
Closed via #758.

PR #758, opened 2016-04-21
#674 Support inheritance in offline compression using a variable

grzegorzswica:
Added support for template inheritance with variable from context in offline compression ([#674](https://github.com/django-compressor/django-compressor/issues/674)).

This required some refactoring:
- `walk_nodes` from `DjangoParser` and `Jinja2Parser` now take additional named parameter - `context`,
- `get_nodelist` from `DjangoParser` takes `context` as parameter,
- offline contexts are now evaluated before nodes discovery in templates,
- node discovery in templates is done for all offline contexts,
- for each compress node found in there is a list of contexts for which the node was present,
- each node is then processed only with relevant contexts.

This change allows for usage of parametrised inheritance of templates:
- directly by passing parent template name in context (test case added),
- indirectly through custom extends tag by passing mocked request in context (no test case provided as it is a special case of the above).

### DIFF
--- a/compressor/management/commands/compress.py
+++ b/compressor/management/commands/compress.py
@@ -2,7 +2,7 @@
 import os
 import sys
 
-from collections import OrderedDict
+from collections import OrderedDict, defaultdict
 from fnmatch import fnmatch
 from importlib import import_module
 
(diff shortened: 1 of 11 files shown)


diox:
Hi, 

This looks good, thanks! Can you please fix conflicts with master so that I can merge it ?

grzegorzswica:
Squashed and rebased.

diox:
Thanks! This is exactly what I had in mind when I filed #674, happy someone stepped up and fixed it for me :)

MounirMesselmeni:
Hi, when will this be released?

karyon:
there's no ETA, but you can use the development version. it should be quite stable.

Issue #680, opened 2015-10-14
COMPRESS_OFFLINE setting clarification


alfetopito:
Hey there!

I spent awhile struggling to get django-compress working together with [whitenoise](http://whitenoise.evans.io) on [Dokku] (http://progrium.viewdocs.io/dokku) (you know, almost Heroku).

I stared for some time at this ticket https://github.com/django-compressor/django-compressor/issues/486#issuecomment-36886332 and tried several things, all without success.

Even disabled whitenoise's caching, and still no luck.

I had come across the [COMPRESS_OFFLINE] (http://django-compressor.readthedocs.org/en/latest/settings/#django.conf.settings.COMPRESS_OFFLINE) setting, but had ignored it since `compress --force` created the needed file, and it states that when it is set, the compressing will be ALSO done offline together with request/response loop. 

I had thought: "If it won't disable online compression and the --force flag works, why bother?"

Turns out I need to set this to `COMPRESS_OFFLINE = True` to finally get it working.

Before setting this flag, the file name set on my html didn't match the one on the file system.

So, I just suggest the docs reflect the need to set that flag if working offline. At least I had to.

diox:
`COMPRESS_OFFLINE` makes `{% compress %}` tag use offline compression at request/response cycle time. It's a specific way of using compressor, entirely optional, as explained in http://django-compressor.readthedocs.org/en/latest/scenarios/

The `compress` management command is only really meant for use with offline compression, that's why you have to use `--force` if offline compression is not enabled. If offline compression is disabled and you run the command with `--force`, that will generate files it would normally generate during the request/response cycle (and the offline manifest, which will not be used, since offline compression is disabled), but with the caveats explained in the doc I linked above. It's usually a bad idea.

Feel free to submit a PR with specific documentation tweaks once you have a good idea of the clarifications you seek - it's not clear to me what you'd want the docs to say at this point (probably because I'm fairly familiar with the way compressor works already).

diox:
Also worth checking out http://django-compressor.readthedocs.org/en/latest/behind-the-scenes/

alfetopito:
> COMPRESS_OFFLINE makes {% compress %} tag use offline compression at request/response cycle time. 

Oh, I see. That makes more sense.

What got me confused was the ALSO. If you have the flag set, does the compression still happens at request time? I assume it doesn't from your response. 

[...]

PR #761, opened 2016-04-27
Minor docs clarification on COMPRESS_OFFLINE.

mrazzari:
From #680

> If you have the flag set, does the compression still happens at request time? 

This was a bit confusing due to "also" in the description implying that should offline fail, the full compress flow would be used as fallback. 
Added a link to further info that clarifies this behavior.

### DIFF
--- a/docs/settings.txt
+++ b/docs/settings.txt
@@ -458,9 +458,9 @@ Offline settings
 
     :Default: ``False``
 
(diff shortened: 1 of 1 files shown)


karyon:
thanks! i'll leave the issue open as i think that part can be improved further by e.g. recommendations on when to use the setting, but this PR should prevent more confused users.

PR #762, opened 2016-04-27
Fixing CssAbsoluteFilter changes double quotes to single quotes which causes svg to fail

### LINKED ISSUE #738: CssAbsoluteFilter changes double quotes to single quotes which causes svg to fail
CssAbsoluteFilter converts

```
url("data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' xmlns:xlink='http://www.w3.org/1999/xlink' x='0px' y='0px' width='25px' height='25px' viewBox='0 0 25 25' enable-background='new 0 0 25 25' xml:space='preserve'><rect id='XMLID_1_' x='4.9' y='10.9' fill='#18235D' width='15' height='3'/></svg>") no-repeat center
```
into this
```
url('data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' xmlns:xlink='http://www.w3.org/1999/xlink' x='0px' y='0px' width='25px' height='25px' viewBox='0 0 25 25' enable-background='new 0 0 25 25' xml:space='preserve'><rect id='XMLID_1_' x='4.9' y='10.9' fill='#18235D' width='15' height='3'/></svg>') no-repeat center
```

and Chrome says the property is invalid. I'm guessing this is because of the single quotes that are in the svg. To work around this problem, I just set
```
COMPRESS_CSS_FILTERS = []
```

intellisense:
Fixes #738

### DIFF
--- a/compressor/filters/css_default.py
+++ b/compressor/filters/css_default.py
@@ -84,19 +84,23 @@ class CssAbsoluteFilter(FilterBase):
 
     def _converter(self, matchobj, group, template):
         url = matchobj.group(group)
-        url = url.strip(' \'"')
+
+        url = url.strip()
+        wrap = '"' if url[0] == '"' else "'"
+        url = url.strip('\'"')
+
         if url.startswith('#'):
-            return "url('%s')" % url
+            return template % (wrap, url, wrap)
         elif url.startswith(SCHEMES):
-            return "url('%s')" % self.add_suffix(url)
+            return template % (wrap, self.add_suffix(url), wrap)
         full_url = posixpath.normpath('/'.join([str(self.directory_name),
                                                 url]))
         if self.has_scheme:
             full_url = "%s%s" % (self.protocol, full_url)
-        return template % self.add_suffix(full_url)
+        return template % (wrap, self.add_suffix(full_url), wrap)
 
     def url_converter(self, matchobj):
-        return self._converter(matchobj, 1, "url('%s')")
+        return self._converter(matchobj, 1, "url(%s%s%s)")
 
     def src_converter(self, matchobj):
-        return self._converter(matchobj, 2, "src='%s'")
+        return self._converter(matchobj, 2, "src=%s%s%s")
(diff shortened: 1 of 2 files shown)


karyon:
thanks! i should've mentioned that right away but unfortunately forgot it, i need a test before including this. basically, something like running the AbsoluteFilter on that svg and ensuring the correct outcome, and then doing the same with " and ' exchanged and ensuring the correct outcome once more.

karyon:
hey, do you think there is any chance i get some tests from you in the next few weeks? :)

intellisense:
@karyon Sorry for the delay, was pretty busy in the work. I will write tests on coming weekend. Thanks!

karyon:
awesome, thanks!

intellisense:
@karyon The URL wrapped with single quotes was already being tested [here](https://github.com/django-compressor/django-compressor/blob/develop/compressor/tests/test_filters.py#L280) so I added a one new test to check URL wrapped with double quotes should return a URL wrapped with double quotes.

During testing I found a new issue, consider this valid CSS which uses SVG URL:

```
body { background-image: url("data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' width='10' height='10'><linearGradient id='gradient'><stop offset='10%' stop-color='%23F00'/><stop offset='90%' stop-color='%23fcc'/> </linearGradient><rect fill='url(%23gradient)' x='0' y='0' width='100%' height='100%'/></svg>");}
```

The [URL_PATTERN](https://github.com/django-compressor/django-compressor/blob/develop/compressor/filters/css_default.py#L9) will choke and resulting in an improper match because it will stop at first closing parentheses `)` and gives:

```
"data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' width='10' height='10'><linearGradient id='gradient'><stop offset='10%' stop-color='%23F00'/><stop offset='90%' stop-color='%23fcc'/> </linearGradient><rect fill='url(%23gradient
```

Instead of returning the full URL till `</svg>"` because if you check the SVG has `fill` property which is also a URL. Should I file a separate issue? (This requires we should not be using regex to match URL pattern because we need to remember all the opening parenthesis or we should handle SVG separately altogether)

Thanks!

karyon:
nice, thanks. yes please file a new issue for that. could you squash all the commits into one? otherwise it looks good.

intellisense:
@karyon Squashed the commits into one. Thanks!

PR #767, opened 2016-05-18
Make it clear how it should look when finished

camilonova:


karyon:
Continued in #771, thanks!

PR #770, opened 2016-06-01
Clarify offline compression

### LINKED ISSUE #680: COMPRESS_OFFLINE setting clarification

karyon:
Fixes #680.

The example for `COMPRESS_OFFLINE_CONTEXT` is quite bad, but the previous one was even worse because of that `path_to_files` that should probably have been in `STATIC_URL`.

### DIFF
--- a/docs/behind-the-scenes.txt
+++ b/docs/behind-the-scenes.txt
@@ -1,6 +1,6 @@
 .. _behind_the_scenes:
 
-Behind the scenes
+Behind the Scenes
 =================
 
 This document assumes you already have an up and running instance of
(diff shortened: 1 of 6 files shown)


diox:
Thanks, that was a much needed update, it looks great!

PR #771, opened 2016-06-01
Remote storages docs: clarify result

karyon:
Expand on #767


karyon:
@camilonova could you confirm again that settings like the ones in this PR work for you? unfortunately, i cannot test this by myself...

camilonova:
@karyon works fine so far. Thanks.

PR #779, opened 2016-07-13
Allowing it to work with offline cache backend

jasonamyers:
Could we consider allowing it to fail in a way that works with the backend cache offline?  This is a very naive implementation. I'd be happy to take it further if others feel it's useful to them. Taking it further in my mind is setting it up to catch the specific ConnectionError type for the supported backends, adding logging, testing, and maybe a setting so you can turn the SilentFail on or off.

Signed-off-by: Jason Myers jason@jasonamyers.com

### DIFF
--- a/compressor/cache.py
+++ b/compressor/cache.py
@@ -97,10 +97,16 @@ def get_templatetag_cachekey(compressor, mode, kind):
 def get_mtime(filename):
     if settings.COMPRESS_MTIME_DELAY:
         key = get_mtime_cachekey(filename)
-        mtime = cache.get(key)
+        try:
+            mtime = cache.get(key)
+        except Exception:
+            mtime = None
         if mtime is None:
             mtime = os.path.getmtime(filename)
-            cache.set(key, mtime, settings.COMPRESS_MTIME_DELAY)
+            try:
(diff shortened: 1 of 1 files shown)


karyon:
sorry for the late reply. to me this seems like an edge use case, when the cache does not exist at all, then the setup is likely to be quite fundamentally broken. in my opinion crashing loudly is the right thing to do. what do you think?

jasonamyers:
I really want my site to stay working if for some reason something happens to my cache resource. I've had a few outages at work that were due to AWS and Rackspace issues that were temporary. I get notified when those occur via my logging, but I wished it didn't stop my users from being able to use the site. Going slower but working is better than a 500 for my use case. I view cache as a bonus, and don't want it to be a detriment to my site. I can understand why you want it to fail loudly. We can just carry a private fork if you don't want this capability.

jasonamyers:
Either way you decide thank you for a wonderful package. I appreciate your work!

diox:
Hi,

I understand the idea but I'd prefer compressor to not handle edge cases like that.

If you want silent fallback, you can implement it yourself in your own caching backend. I've done that in the past, it's not very hard, just dangerous, you need to know what you are doing :)

jasonamyers:
:+1:

PR #795, opened 2016-09-06
Update index.txt

barseghyanartur:
Fix typo.


karyon:
i think this is intentional :)

image

Issue #736, opened 2016-01-29
Compression with CSS filters break image file name with space in CSS result


stasdavydov:
I have the background image CSS rule used file with a space in the name:

```
#checkout_form .credit-cards {
    background-image: url("../imgs/Credit cards.png");
}
```

{% compress css inline %} block produce the following result:
`#checkout_form .credit-cards{background-image: url("../imgs/Creditcards.png")}`
Space in file name is removed.

I use the following CSS filters:
`COMPRESS_CSS_FILTERS = ['compressor.filters.css_default.CssAbsoluteFilter',  'compressor.filters.cssmin.CSSMinFilter']`

karyon:
hey, it would be cool if you could check whether it's the AbsoluteFilter or the minifier!

karyon:
So, it looks like rcssmin is causing this. i built a test to assert the CssAbsoluteFilter is working correctly (#805) and opened https://github.com/ndparker/rcssmin/issues/7.

PR #805, opened 2016-10-30
Test spaces in css urls

karyon:
Add a test to show that #736 is not caused by the CssAbsolute filter but by https://github.com/ndparker/rcssmin/issues/7 instead.


compressor/tests/test_filters.py
```diff
@@ -234,12 +234,6 @@ def test_css_no_hash(self):
         filter = CssAbsoluteFilter(content)
         self.assertEqual(output, filter.input(filename=filename, basename='css/url/test.css'))
 
-        with self.settings(COMPRESS_URL='http://static.example.com/'):
```
diox: I would not consider that duplication. Testing with `COMPRESS_URL` being an absolute URL has merit. Same for similar instances below.

karyon: look again! :) there's a new test class below

diox: Totally missed that, it's fine, carry on

Issue #778, opened 2016-07-06
Make offline compression independent of STATIC_URL


kottenator:
I use offline compression. I do `manage.py compress` once and then deploy the project with gathered static files to multiple servers. 

On different servers I can have different `STATIC_URL`, not the one used for compression. That's problematic for current `django-compressor` implementation because:
- it calculates _offline manifest keys_ using _actual_ `STATIC_URL` when in the `manifest.json` it corresponds to `STATIC_URL` used during `manage.py compress`. And it raises an exception - "the key is not found in the offline manifest"
- another problem is that in offline manifest you store pieces of HTML that also contain _obsolete_ `STATIC_URL` which doesn't work on deployed site instance because `STATIC_URL` is different now

You could change 3 things to make it work:
- when calculating manifest keys: substitute all `STATIC_URL` entries with some constant, e.g.: `/static/` → `$STATIC_URL`
- when generating offline manifest via `manage.py compress` - don't put actual `STATIC_URL` (e.g. `/static/`) into `manifest.json` but put some constant like `$STATIC_URL` instead
- when reading the manifest - replace all `$STATIC_URL` with actual `STATIC_URL`, e.g.: `$STATIC_URL` → `/new/static/`

kottenator:
I've created [a gist](https://gist.github.com/kottenator/fe4b0a05d1645af6986fbfdd407c57a7) that _patches_ `django-compressor` - it's a quick-and-dirty solution for my project. Maybe you'll find it useful.

karyon:
may i ask, why do you have different `STATIC_URL`s?

kottenator:
Sure, for example, I host my Django app on multiple sub-paths of the same domain, like:

* http://example.com/dev/my-app/
* http://example.com/staging/my-app/

Unfortunately, we can't use sub-domains for that. So the static files are hosted at:

* http://example.com/dev/my-app/static/
* http://example.com/staging/my-app/static/

I configure `STATIC_URL` via environment variable, so it's `/dev/my-app/static/` and `/staging/my-app/static/` correspondingly.

But `django-compressor` has generated static files with `STATIC_URL = '/static/'`.

diox:
Can't you use a different offline manifest for both environments ?

kottenator:
I build Django project distribution once and deploy it on multiple instances. The dist doesn't know about where it'll be deployed - it's a matter of configuration.

kottenator:
[...]

PR #809, opened 2016-11-21
Make offline compression and decompression independent of STATIC_URL

kottenator:
If `settings.COMPRESS_OFFLINE_URL_PLACEHOLDERS` is `True` then:
1. Replace `settings.COMPRESS_URL` with `settings.COMPRESS_URL_PLACEHOLDER` in the offline manifest when doing offline compression via `manage.py compress`
2. Calculate offline manifest keys with `settings.STATIC_URL`replaced with `''`

In this way we can do offline compression with one `settings.STATIC_URL` and then deploy it with different `settings.STATIC_URL` - `manifest.json` will work with any static URL.

More details are in #778.


karyon:
unfortunately i don't know enough of the code to see whether there might be a better way. right now this feels rather bolted on top. @kottenator have you thought about any other ways? @mat any thoughts on this?

diox:
Sorry, I don't really like the approach. Altering settings at run-time is a bad idea IMHO, and doing it for a pretty uncommon use-case doesn't seem worth it.

kottenator:
The problem is in Compressor approach - it renders the node contents, which often are `{% static %}` tags, which use `settings.STATIC_URL`. 

There is no better way to implement the URL independence.

kottenator:
And also - I've made this optional, and only for offline compression.

kottenator:
Well, actually there may be a way without modifying `settings`, but it would be something like:

* render `{% compress %}` node
* replace all `STATIC_URL` entries with placeholder
* calculate the manifest key

kottenator:
It was easy to change :)
@diox, @karyon - please, take a look again.

compressor/management/commands/compress.py
```diff
@@ -250,7 +250,15 @@ def compress(self, log=None, **options):
...
-                    offline_manifest[key] = result
+
+                    if settings.COMPRESS_OFFLINE_USE_URL_PLACEHOLDER:
+                        # Replace settings.COMPRESS_URL with a placeholder
+                        offline_manifest[key] = result.replace(
+                            settings.COMPRESS_URL, settings.COMPRESS_URL_PLACEHOLDER
+                        )
+                    else:
+                        offline_manifest[key] = result
+
```
karyon: i was wondering, why don't you do `result = result.replace(....)` as you do in `templatetags/compress.py `? this makes this method return results that did not get written to the manifest.

looking at this i noticed that this `compress` method writes the manifest, and the `handle` method reads that manifest, merges it with others, and writes again. wouldn't it be better to not write the manifest here, just return it, and let `handle` do the writing? @mat?

compressor/tests/test_offline_url_placeholders.py
```diff
@@ -0,0 +1,59 @@
...
+
+for key in dir(compressor.tests.test_offline):
+    value = getattr(compressor.tests.test_offline, key)
+    if (
+        key not in ('OfflineCompressStaticTemplateTagTestCase', 'TestCompressCommand') and
+        # Find all the TestCase subclasess except the TestCase itself
+        isinstance(value, type) and issubclass(value, TestCase) and value is not TestCase
+    ):
+        new_class = type(key, (value,), {})
+        module_dict[key] = override_settings(COMPRESS_OFFLINE_USE_URL_PLACEHOLDER=True)(new_class)
```
karyon: what does all this do?

compressor/contrib/jinja2ext.py
```diff
@@ -45,11 +45,10 @@ def _compress_normal(self, kind, mode, caller):
...
-        original_content = caller()
         context = {
-            'original_content': original_content
+            'original_content_generator': caller,
         }
         return self.render_compressed(context, kind, mode, forced=forced)
 
     def get_original_content(self, context):
-        return context['original_content']
+        return context['original_content_generator']()
```
karyon: these seem unrelated...?

compressor/management/commands/compress.py
karyon: sorry, @diox :)

karyon:
added some random comments, didn't do an actual review. i'd wait for @diox opinion on the general approach (and the offline manifest writing thing i commented on)

compressor/management/commands/compress.py
kottenator: Yes, I see what you mean: writing the manifest, then reading, then writing again. 
I'm not sure though if it'll give the same effect if I'll move it to `.handle()`...

compressor/tests/test_offline_url_placeholders.py
kottenator: There is `compress/tests/test_offline.py` file, it has ~60 tests. I want to re-use them: to run them all but with `settings.COMPRESS_OFFLINE_USE_URL_PLACEHOLDER=True`.

I'm dynamically creating subclasses with overridden setting.

compressor/contrib/jinja2ext.py
kottenator: I wanted to ask you about this.

In Django you always re-render the node contents but in Jinja you render it once and then re-use. This has brought some problems when I was overriding Django settings. Now it doesn't, but shouldn't it be implemented in the same way for Django and Jinja?

karyon: i don't know :) but @diox might...? if you can write a test that demonstrates the problem, i could merge this, but i'd also like to have this in a separate commit.

compressor/tests/test_offline_url_placeholders.py
karyon: okay... provided @diox signs off the general approach of this PR, this part here definitely needs a comment explaining exactly that.

compressor/contrib/jinja2ext.py
kottenator: Yes, probably it's a separate thing. I'll remove it now.

kottenator: Reverted.

compressor/management/commands/compress.py
karyon: for the record the double write was added in https://github.com/django-compressor/django-compressor/pull/806

kottenator:
@diox - I've made all (except one) the change requests and have removed `settings` modification. Please, review this PR again, when you'll have a chance.

diox:
I haven't got the time to do a proper review, but I like this approach a lot better. The manipulation to run all the tests is kinda hairy, but I suppose you don't have much choice with the way they are written (if I could do it all over again I'd do a single class).

We'll need docs about the new settings.

kottenator:
Sure, I'll add the docs.

kottenator:
Docs added.

compressor/management/commands/compress.py
karyon: huh, even your own test uses the return values. i don't get it. i hope you can enlighten me :)

docs/settings.txt
```diff
@@ -516,3 +516,21 @@ Offline settings
...
+
+.. attribute:: COMPRESS_OFFLINE_URLLESS
+
+    :Default: ``False``
+
+    Enable URL-less mode when you do offline compression & decompression regardless of
+    ``settings.STATIC_URL`` and ``settings.COMPRESS_URL`` so your assets become re-locatable
+    for any URLs.
+
+    Affects manifest file contents and manifest key determination, but not the assets contents.
```
karyon: "for any URL" and "asset's"

docs/settings.txt
```diff
@@ -516,3 +516,21 @@ Offline settings
...
+
+.. attribute:: COMPRESS_URL_PLACEHOLDER
+
+    :Default: ``'/__compressor_url_placeholder__/'``
+
+    If :attr:`~django.conf.settings.COMPRESS_OFFLINE_URLLESS` is ``True`` -
+    replace :attr:`~django.conf.settings.COMPRESS_URL` with this placeholder in the
+    offline manifest file.
```
karyon:
```
    If :attr:`~django.conf.settings.COMPRESS_OFFLINE_URLLESS` is ``True``,
    each occurrence of :attr:`~django.conf.settings.COMPRESS_URL` in the offline
    manifest file will be replaced with this placeholder.
```

compressor/management/commands/compress.py
```diff
@@ -250,7 +250,15 @@ def compress(self, log=None, **options):
                     except Exception as e:
                         raise CommandError("An error occurred during rendering %s: "
                                            "%s" % (template.template_name, smart_text(e)))
-                    offline_manifest[key] = result
+
+                    if settings.COMPRESS_OFFLINE_URLLESS:
+                        # Replace settings.COMPRESS_URL with a placeholder
```
karyon: as below, remove comment

docs/settings.txt
kottenator: Done.

docs/settings.txt
kottenator: Done.

compressor/templatetags/compress.py
```diff
@@ -64,7 +64,15 @@ def render_offline(self, context):
         key = get_offline_hexdigest(self.get_original_content(context))
         offline_manifest = get_offline_manifest()
         if key in offline_manifest:
-            return offline_manifest[key]
+            result = offline_manifest[key]
+
+            if settings.COMPRESS_OFFLINE_URLLESS:
+                # Replace placeholder with original settings.COMPRESS_URL
```
kottenator: Done.

compressor/conf.py
```diff
@@ -70,6 +70,12 @@ class CompressorConf(AppConf):
     OFFLINE_MANIFEST = 'manifest.json'
     # The Context to be used when TemplateFilter is used
     TEMPLATE_FILTER_CONTEXT = {}
+    # URL-less mode: you do offline compression & decompression regardless of settings.STATIC_URL
+    # and settings.COMPRESS_URL so your assets become re-locatable for any URLs.
+    # Affects manifest file contents and manifest key determination, but not the assets contents.
+    OFFLINE_URLLESS = False
+    # Placeholder to be used instead of settings.COMPRESS_URL while offline compression
```
kottenator: Done.

compressor/management/commands/compress.py
kottenator: _I've replied but now I've lost the reply, so I'll reply again_ :)

I agree, that's a controversial part. My idea was:

* `Command.compress()` returns full functional HTML code so you can use it
* when `manifest.json` contains _unprocessed_ (technical) code that should be _processed first_ before usage

If it's too controversial - I could try to simplify it.

compressor/management/commands/compress.py
kottenator: Done.

kottenator: Done.

compressor/cache.py
```diff
@@ -50,6 +50,9 @@ def get_mtime_cachekey(filename):
 
 
 def get_offline_hexdigest(render_template_string):
+    if settings.COMPRESS_OFFLINE_URLLESS:
+        # Make hexdigest calculation independent of settings.STATIC_URL
+        render_template_string = render_template_string.replace(settings.STATIC_URL, '')
```
kottenator: BTW, maybe not in this PR, but how do you like the idea - to always do this?

compressor/conf.py
```diff
@@ -70,6 +70,12 @@ class CompressorConf(AppConf):
     OFFLINE_MANIFEST = 'manifest.json'
     # The Context to be used when TemplateFilter is used
     TEMPLATE_FILTER_CONTEXT = {}
+    # URL-less mode: you do offline compression & decompression regardless of settings.STATIC_URL
+    # and settings.COMPRESS_URL so your assets become re-locatable for any URLs.
+    # Affects manifest file contents and manifest key determination, but not the assets contents.
+    OFFLINE_URLLESS = False
```
karyon: in my opinion the key determination is a technical detail that's done purely to make the placeholder thing work.... @diox any thoughts on this?

compressor/management/commands/compress.py
kottenator: Let me try to simplify it then...

compressor/conf.py
kottenator: IMO, the placeholder is a technical detail of `STATIC_URL`-independent mode implementation :) 

Nobody should care how is it implemented: using placeholders or by generating manifest in absolutely different way.

compressor/management/commands/compress.py
kottenator: Ok, I've done what you had asked. Now `Command.compress()` returns exactly what is stored in the manifest.

I had to refactor `test_offline.py` to make it _reusable_ for this case.

compressor/conf.py
diox: My 2 cents:
- `OFFLINE_URLLESS=True` should just be the default behavior and not exposed as a setting.
- `URL_PLACEHOLDER` doesn't technically need to be exposed as a setting.

karyon: @diox just to make sure i get this right, you propose to remove the URLLESS setting altogether and just always do the replacement with no option to turn it off? as in, if we had seen earlier that this approach is cool, most of @kottenator's work wouldn't have been necessary ;) ? (i think we should do it this way, just am a bit annoyed that i/we didn't think of it earlier)

@kottenator can you think of any downsides?

karyon: i'd say in this case `URL_PLACEHOLDER` can still be a setting because that's a good place for global constants (or is there a good alternative?), but not document it because people should generally not care about it.

diox: Yes. Sorry for the back-and-forth, but I think it's worth it to reduce complexity. For the record I still think the original issue is not a very common use-case, but after thinking about this for a while I don't see much downside in always doing the replace.

diox: > i'd say in this case URL_PLACEHOLDER can still be a setting because that's a good place for global constants (or is there a good alternative?), but not document it because people should generally not care about it.

Fair enough.

compressor/cache.py
```diff
@@ -50,7 +50,7 @@ def get_mtime_cachekey(filename):
 
 
 def get_offline_hexdigest(render_template_string):
-    return get_hexdigest(render_template_string)
+    return get_hexdigest(render_template_string.replace(settings.STATIC_URL, ''))
```
diox: Add a comment explaining why

compressor/tests/test_offline.py
```diff
@@ -72,10 +72,8 @@ def setUp(self):
         # TEMPLATES[1] to be Jinja2 templates backend in test_settings.
         TEMPLATES = copy.deepcopy(settings.TEMPLATES)
 
-        django_template_dir = os.path.join(
-            TEMPLATES[0]['DIRS'][0], self.templates_dir)
-        jinja2_template_dir = os.path.join(
-            TEMPLATES[1]['DIRS'][0], self.templates_dir)
+        django_template_dir = os.path.join(TEMPLATES[0]['DIRS'][0], self.templates_dir)
+        jinja2_template_dir = os.path.join(TEMPLATES[1]['DIRS'][0], self.templates_dir)
```
diox: Why change this ? It's now breaking 79 chars limit, which was the reason for the current wrapping (I know all compressor code doesn't follow pep8, but let's not regress on the files that do)

diox: Same for a lot of changes below.

kottenator: Sure, I'll bring it back.

compressor/cache.py
kottenator: OK, will do now.

compressor/tests/test_offline.py
kottenator: I'm just using 99 chars limit, which is [allowed by PEP8](https://www.python.org/dev/peps/pep-0008/#maximum-line-length). But I don't want to break your rules.

compressor/management/commands/compress.py
```diff
@@ -246,7 +246,9 @@ def compress(self, log=None, **options):
                         continue
 
                     try:
-                        result = parser.render_node(template, context, node)
+                        result = parser.render_node(template, context, node).replace(
+                            settings.COMPRESS_URL, settings.COMPRESS_URL_PLACEHOLDER
```
karyon: please do the replace outside the try block

compressor/tests/test_offline.py
```diff
@@ -364,6 +357,18 @@ class OfflineCompressStaticTemplateTagTestCase(OfflineTestCaseMixin, TestCase):
...
+        count, result = CompressCommand().compress(
+            log=self.log, verbosity=self.verbosity, engine=engine
+        )
+        self.assertEqual(1, count)
+        self.assertEqual([self._render_script(self.expected_hash)], result)
+        self.assertEqual(self._render_template(engine), self._render_result(result))
+
+        # Changing settings.STATIC_URL doesn't affect the next "offline" decompression
+        with self.settings(STATIC_URL='/another/static/url/'):
+            self.assertEqual(self._render_template(engine), self._render_result(result))
```
karyon: could you make it more explicit what is being tested here? i don't fully understand the structure of the tests but my feeling is this should go in a new Class whose name roughly describes what's being tested. 

also, could you clarify what you mean by "decompression" in the comment? we currently don't use that word anywhere

compressor/tests/test_offline.py
kottenator: Done, 79 chars limit is back to this file (and even better than it was).

compressor/tests/test_offline.py
```diff
@@ -364,6 +376,21 @@ class OfflineCompressStaticTemplateTagTestCase(OfflineTestCaseMixin, TestCase):
...
+        self.assertEqual(1, count)
+        self.assertEqual([self._render_script(self.expected_hash)], result)
+        self.assertEqual(
+            self._render_template(engine), self._render_result(result))
+
+        # Changing settings.STATIC_URL doesn't affect the next
+        # offline "decompression"
+        with self.settings(STATIC_URL='/another/static/url/'):
+            self.assertEqual(
+                self._render_template(engine), self._render_result(result))
```
karyon: that comment got hidden for some reason so here it is again:



could you make it more explicit what is being tested here? i don't fully understand the structure of the tests but my feeling is this should go in a new Class whose name roughly describes what's being tested.

also, could you clarify what you mean by "decompression" in the comment? we currently don't use that word anywhere

compressor/management/commands/compress.py
kottenator: Done.

compressor/tests/test_offline.py
kottenator: Sure, I'll do a separate test class.

By "offline decompression" I mean extraction & processing of the offline manifest content into the HTML output.

compressor/tests/test_offline.py
kottenator: Sure, I'll do a separate test class.

By "offline decompression" I mean:

* offline manifest key determination
* extraction & processing of the offline manifest content into the HTML output

compressor/tests/test_offline.py
```diff
@@ -471,6 +460,40 @@ class OfflineCompressTestCaseWithContextGeneratorSuper(
...
+
+    We use COMPRESS_OFFLINE_CONTEXT generator to make sure that
+    STATIC_URL is not cached when rendering the template.
+    """
+    templates_dir = 'test_static_url_independence'
+    expected_hash = '44ed960a3d1d'
+    additional_test_settings = {
+        'STATIC_URL': '/custom/static/url/',
+        'COMPRESS_OFFLINE_CONTEXT': (
+            'compressor.tests.test_offline.static_url_context_generator'
```
karyon: did you want to specifically use the offline context thing? if not, you could just use the static tag in the template and wouldn't need the offline context

karyon: ...okay i see. you can still use a dictionary here instead of the generator, that'll be a bit less code.

kottenator: The problem with `{% static %}` tag is that it has _cached_ storage.

OK, I'll try to use a dict.

kottenator: No, unfortunately `dict` is also getting cached because it's calculated only once, during tests _compilation_.

karyon:
@kottenator you're awesome for being so patient! :) and thank you for your contribution.

@diox i saw your two review comments addressed and assumed there's nothing more. if yes, i can fix that.

kottenator:
Thank you, @karyon and @diox! Keep your framework being awesome! ;)

PR #810, opened 2016-11-28
Feature: CSS relative URL filter

kottenator:
Do a similar to `CssAbsoluteFilter` URL processing but replace `settings.COMPRESS_URL` prefix with `'../../'`.

## Why do I need it?

The advantage of `CssRelativeFilter` is that we don't hard-code `STATIC_URL` in the bundle code thus we can host the bundle at any `STATIC_URL`.

## How does it work?

For example:

We have `background: url(images/image.png)` in `myapp/static/myapp/styles.css`. 
After compression, `myapp/styles.css` becomes a part of `CACHE/css/abcdef123456.css`.
`CssRelativeFilter` thansforms the image URL inside CSS into `../../myapp/images/image.png`.
The bundle is served at `/static/CACHE/css/abcdef123456.css` so the image URL will be calculated as `/static/CACHE/css/` + `../../myapp/images/image.png` = `/static/myapp/images/image.png`.

## Misc

`CssRelativeFilter` is an alternative to `CssAbsoluteFilter` - you can't use them both.

And as an alternative to `CssAbsoluteFilter`, it's enabled for precompilers when `COMPRESS_ENABLED = False`.


kottenator:
@diox, @karyon - please, take a look. I need this change together with #809 to have full independence of `STATIC_URL`.

karyon:
theoretically, can you think of any downside to making this new filter the default?

i noticed there is a setting `COMPRESSOR_OUTPUT_DIR` which defaults to `"CACHE"`. so i'd guess that users could set that to `some/longer/path` and your assumption that you can go exactly two levels up would break. also, documentation is missing. but let's first discuss the approach before you invest more time into this.

i need to re-understand what the absolute filter is doing and then understand what you are doing and why, will do that sometime later, i don't have the time right now.

also, if you really actually need this change right now, you can configure pip to install compressor from your own fork - but i guess you knew that already :)

kottenator:
@karyon - thanks for the reply! Sure, take a look when you'll have time. Meanwhile I'll answer to some of your questions.

> theoretically, can you think of any downside to making this new filter the default?

The only thing that I can imagine - if somebody has a similar problem to mine, but with `COMPRESS_OUTPUT_DIR` - they need it to be flexible, when `STATIC_URL` is fixed.

> ... users could set that to `some/longer/path` and your assumption that you can go exactly two levels up would break.

That's right, it'll break my filter. Thanks for pointing it out. But it's easy to fix and I've actually done this already :)

> ... also, documentation is missing.

Yes, I've added it now.

> ... but let's first discuss the approach before you invest more time into this.

Sorry, I've already done some changes :) But sure, let's discuss.

kottenator:
@karyon, @diox - did you have a chance to review this PR (again)?

karyon:
i haven't forgotten this, i'm just uncertain what to do since i don't know how to decide whether this makes sense. i poked @diox, maybe he can help.

> The only thing that I can imagine - if somebody has a similar problem to mine, but with COMPRESS_OUTPUT_DIR - they need it to be flexible, when STATIC_URL is fixed.

could you elaborate? didn't quite get this.

compressor/filters/css_default.py
```diff
@@ -104,3 +104,28 @@ def url_converter(self, matchobj):
...
+        new_prefix = ['..']
+        # N levels up from settings.COMPRESS_OUTPUT_DIR:
+        # - if it's 'CACHE' (default) then it'd be ['..'];
+        # - if it's 'CACHE/in/depth' then it'd be ['..', '..', '..'].
+        new_prefix += [
+            '..' for part in os.path.normpath(
+                settings.COMPRESS_OUTPUT_DIR
+            ).split(os.sep) if part
+        ]
+        new_prefix = '/'.join(new_prefix)
```
karyon: maybe something like
```
# One level up from 'css' / 'js' folder
new_prefix = '..'
# N levels up from settings.COMPRESS_OUTPUT_DIR
levels = len(os.path.normpath(settings.COMPRESS_OUTPUT_DIR).split(os.sep))
new_prefix += (os.sep + '..') * levels
```
the `if part` shouldn't be necessary when using normpath, or did you add it for a specific reason?

compressor/filters/css_default.py
```diff
@@ -104,3 +104,28 @@ def url_converter(self, matchobj):
...
         return self._converter(matchobj, 2, "src=%s%s%s")
+
+
+class CssRelativeFilter(CssAbsoluteFilter):
+    """
+    Do similar to ``CssAbsoluteFilter`` URL processing
+    but replace ``settings.COMPRESS_URL`` prefix with '../../'.
+    """
+
+    def add_suffix(self, url):
```
karyon: ... a method called add_suffix doesn't seem to be the right place for this. maybe you find something better?

compressor/filters/css_default.py
```diff
@@ -104,3 +104,28 @@ def url_converter(self, matchobj):
 
     def src_converter(self, matchobj):
         return self._converter(matchobj, 2, "src=%s%s%s")
+
+
+class CssRelativeFilter(CssAbsoluteFilter):
+    """
+    Do similar to ``CssAbsoluteFilter`` URL processing
+    but replace ``settings.COMPRESS_URL`` prefix with '../../'.
```
karyon: the `'../../'` part is not correct anymore

docs/settings.txt
```diff
@@ -87,6 +87,12 @@ Backend settings
          feature, and the ``'content'`` in case you're using multiple servers
          to serve your content.
 
+    - ``compressor.filters.css_default.CssRelativeFilter``
+
+      Alternative to ``CssAbsoluteFilter`` which uses ``'../../'``
```
karyon: needs updating as well

kottenator:
> The only thing that I can imagine - if somebody has a similar problem to mine, but with `COMPRESS_OUTPUT_DIR` - they need it to be flexible, when `STATIC_URL` is fixed.

> @karyon:: Could you elaborate? didn't quite get this.

I mean that in my case I have flexible `STATIC_URL`:

* which is `/a/static/` when I generate the assets
* and it's `/b/static/` when I deploy the assets

Maybe someone has similar situation with `COMPRESS_OUTPUT_DIR`:

* it's `CACHE` when they generate the assets
* and its `data/compressed` when they deploy the assets

Though maybe I'm wrong and it'll not break anything.

compressor/filters/css_default.py
kottenator: Sure, I'll update the comment.

docs/settings.txt
kottenator: Sure, I'll update the comment.

compressor/filters/css_default.py
kottenator: In general - looks better, thanks! I'll use your code ;)

> the `if part` shouldn't be necessary when using `normpath`

Due to `django-compressor` implementation, the following cases will be treated the same:

* `settings.COMPRESS_OUTPUT_DIR = 'CACHE'
* `settings.COMPRESS_OUTPUT_DIR = '/CACHE'

So `normpath` doesn't solve the problem. Unless I'll do some conditional replacement.

`(os.sep + '..')` is wrong; `new_prefix` should be a URL, not file path.

compressor/filters/css_default.py
kottenator: Comment updated.

docs/settings.txt
kottenator: Comment updated.

compressor/filters/css_default.py
kottenator: The code updated.

compressor/filters/css_default.py
kottenator: I agree, it looks strange at glance. But I don't see a better place.

I started with the idea to override `self.input()` method. But then I realized that after `output = super().input(...)` I'll get already processed URLs so it'll be hard to detect which were _related initially_.

So I need to override some _atomic_ method that processes one URL.

There are two methods that do that:

```py
class CssAbsoluteFilter(FilterBase):
    def url_converter(self, matchobj):
        return self._converter(matchobj, 1, "url(%s%s%s)")

    def src_converter(self, matchobj):
        return self._converter(matchobj, 2, "src=%s%s%s")
    # ...
```

Method `self._converter()` does a bunch of useful things that I don't want to copy-paste, and finally, it calls `self.add_suffix()`.

`self.add_suffix()` means "add a suffix to a relative URL" which is `settings.COMPRESS_URL` in `CssAbsoluteFilter` and `'../' * (N + 1)` in `CssRelativeFilter`.

So probably it's not the worse method to override :)

kottenator:
@karyon - actually, I don't mind to replace `CssAbsoluteFilter` with `CssRelativeFilter` if you and @diox think it's a good idea ;)

But keeping them both is also not so bad idea, IMO.

kottenator:
@karyon - have I satisfied all your change requests or do I need to do something else?

karyon:
the CssAbsoluteFilter changed a bit here: https://github.com/django-compressor/django-compressor/pull/828. please check whether there are functional changes needed on your side, and also there are merge conflicts.

i still disagree with replacing a prefix in a method called `add_suffix`. while there might be technical reasons for this, the contents of the method simply don't match its name. please find a different solution. one idea might be to add and call a new, empty method in the CssAbsoluteFilter and override that one in your filter.

karyon:
ping :)

kottenator:
I'll do your final request(s) today, @karyon.

kottenator:
@karyon: as I've promised - today I've implemented requests from your last comment:

> the CssAbsoluteFilter changed a bit here: #828.

Indeed, there were merge conflicts but I've handled them.

> i still disagree with replacing a prefix in a method called `add_suffix` ... one idea might be to add and call a new, empty method in the `CssAbsoluteFilter` and override that one in your filter.

... and that's what I've done :)

kottenator:
@karyon, @diox - ping ;)

compressor/filters/css_default.py
kottenator: I've refactored the code. Now it's not inside the `.add_suffix()` but in a separate method.

karyon:
thanks!

kottenator:
Thank you, @karyon.

PR #812, opened 2016-12-08
silently ignore link tags if missing rel atrribute

bobthekingofegypt:
Stop compress tag from throwing KeyError when link tag is missing rel attribute

### DIFF
--- a/compressor/css.py
+++ b/compressor/css.py
@@ -16,7 +16,7 @@ class CssCompressor(Compressor):
             data = None
             elem_name = self.parser.elem_name(elem)
             elem_attribs = self.parser.elem_attribs(elem)
-            if elem_name == 'link' and elem_attribs['rel'].lower() == 'stylesheet':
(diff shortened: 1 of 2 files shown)


karyon:
is it allowed to omit the rel thing? or to put it differently, why do you thing compressor should ignore it if it's missing and not throw some error?

bobthekingofegypt:
According to the living spec a link should have either a rel or a itemprop attribute, so it is possible but contrived.  Never having made a custom tag before I just read the django docs that said template tags should fail silently.

I only added this because I was getting a KeyError when I had wrapped a bunch of link tags that chrome didn't have an issue with.  Admittedly it did identify a mistake in the page I was converting.  Maybe an error message saying all link tags wrapped in a compress should have a rel attribute is a better option.

karyon:
as far as i can see, it is valid to not use a rel attribute if this itemprop thing is present. i think compressor shouldn't fail in that case, and i think your solution is the right way to go.

here's a related SO post with links to the specs: http://stackoverflow.com/a/31621309/4312944

thanks!

PR #813, opened 2016-12-30
Fixes a JS syntax error when certain files are combined

vegitron:
compressor will turn files that look like

File1.js:
(function($) { console.log("run"); })(window)

File2.js:
(function($) { console.log("run 2"); })(window)

Into:
(function($) { console.log("run"); })(window)
(function($) { console.log("run 2"); })(window)

which will give an error like "Uncaught TypeError: (intermediate value)(...) is not a function"

This commit changes that output to:

(function($) { console.log("run"); })(window);(function($) { console.log("run 2"); })(window)

### DIFF
--- a/compressor/js.py
+++ b/compressor/js.py
@@ -50,3 +50,20 @@ class JsCompressor(Compressor):
                     ret.append(subnode.output(*args, **kwargs))
                 return '\n'.join(ret)
         return super(JsCompressor, self).output(*args, **kwargs)
+
+    def filter_input(self, forced=False):
+        """
+        Passes each hunk (file or code) to the 'input' methods
+        of the compressor filters.
+        """
+        content = []
+        for hunk in self.hunks(forced):
+            # If a file ends with a function call, say, console.log()
+            # but doesn't have a semicolon, and the next file starts with
+            # a (, the individual files are ok, but when combined you get an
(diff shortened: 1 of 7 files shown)


karyon:
can't we simply put a semicolon in front of every single js file? would make the code much simpler.

compressor/js.py
```diff
@@ -50,3 +50,20 @@ def output(self, *args, **kwargs):
...
+        of the compressor filters.
+        """
+        content = []
+        for hunk in self.hunks(forced):
+            # If a file ends with a function call, say, console.log()
+            # but doesn't have a semicolon, and the next file starts with
+            # a (, the individual files are ok, but when combined you get an
+            # error like TypeError...
+            # Forcing a semicolon in between fixes it.
+            if settings.COMPRESS_ENABLED or forced:
```
vegitron: @karyon OK - changed to just add ";" before each file.  The if is needed so semicolons aren't added before <script> tags when compression is disabled.  I'm not 100% confident in that 2-part test - does that seem reasonable?

karyon: the `settings.COMPRESS_ENABLED or forced`? that seems reasonable to me. or did you mean something else?

karyon:
thanks!

PR #814, opened 2017-01-04
Add more information to error message when importing a class fails

davgibbs:
Hi,

I was hoping to add this change which adds more information to the error message when the "get_class" method raises an exception. This can be useful when trying to debug why a class is not importing correctly.

In my case, I was trying to create source maps with django-compressor using uglify by adding this file here:
https://github.com/roverdotcom/django-compressor-sourcemaps/blob/master/uglify.py

However I was getting an import error on starting Django as the file in the import line "from compressor.utils.stringformat import FormattableString as fstr" no longer exists. So eventually, I found that I needed to update the uglify.py file code. But it would be nice if the exact reason for the import failure was shown in the log. I think this change will add this detail.

Let me know what you think,
Thanks!

### DIFF
--- a/compressor/tests/test_utils.py
+++ b/compressor/tests/test_utils.py
@@ -5,6 +5,8 @@ import django.contrib.staticfiles.finders
 import django
 
 import compressor.utils.staticfiles
+from compressor.exceptions import FilterError
+from compressor.utils import get_class
 
 from imp import reload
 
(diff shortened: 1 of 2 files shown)


compressor/tests/test_utils.py
```diff
@@ -44,3 +46,15 @@ def test_finders_is_none_if_staticfiles_is_not_installed(self):
...
+class TestGetClass(TestCase):
+
+    def test_get_class(self):
+        try:
+            get_class('common.uglify.JsUglifySourcemapCompressor')
+        except FilterError as e:
+            self.assertTrue(('Failed to import common.uglify.JsUglifySourcemapCompressor. '
+                             'ImportError is: No module named' in str(e)))
+        else:
+            self.fail('FilterError not raised')
```
karyon: check out this [stackoverflow answer](http://stackoverflow.com/a/3166985/4312944) which should make this prettier.

davgibbs: Hi @karyon , thanks. I have updated the pull request now to make use of assertRaises()

Issue #438, opened 2013-09-04
Support for js source maps


polmuz:
Source maps are a very nice tool for debugging something on compressed code using Chrome Dev Tools. There are lots of tools that generate them including Google closure that is already supported by django-compressor.

Here's some info: https://github.com/ryanseddon/source-map/wiki/Source-maps%3A-languages,-tools-and-other-info

vladlep:
+1 I would like this feature also. It is very handy to be able to debug JS compressed files in production. Sentry errors do not make much sense if not.

Woile:
+1 for the feature, absolutely required for Sentry errors :)

wtayyeb:
+1 for js and css source maps

ace-han:
I would like to see  js/ES6/coffee(even TypeScript) and css/sass/less source map
🍻

travijuu:
Does Compress still not support source maps feature?

diox:
No, it still does not support that feature. You're welcome to submit a PR implementing this.

imposeren:
It's still possible to implement some "hackish" solution, for example:
https://bitbucket.org/imposeren/tabletop-planning/src/7d551806cfee2c5493b7273e40c5448aec480544/tabletop_tools/compressor_filters.py?at=master

Better solution is required to make PR...

Andrioden:
For esbuild you can use inline sourcemap, and it works with django-compressor

``` js
await esbuild.build({
    // ...
    sourcemap: "inline",
})
```

PR #815, opened 2017-01-11
Add sourcemaps on ClosureCompilerFilter

ciotto:
In order to use [opbeat-plain-JS](https://github.com/opbeat/opbeat-plain-js) to track JavaScript errors, I added a few features for sourcemaps files creation using the ClosureCompilerFilter.

I also add some test for this part.

### DIFF
--- a/compressor/base.py
+++ b/compressor/base.py
@@ -289,16 +289,21 @@ class Compressor(object):
         any custom modification. Calls other mode specific methods or simply
         returns the content directly.
         """
-        output = '\n'.join(self.filter_input(forced))
 
-        if not output:
+        outputs = self.filter_input(forced)
+        if settings.COMPRESS_OFFLINE_GROUP_FILES:
+            outputs = ['\n'.join(outputs)]
+
+        if not outputs:
             return ''
 
         if settings.COMPRESS_ENABLED or forced:
-            filtered_output = self.filter_output(output)
-            return self.handle_output(mode, filtered_output, forced)
-
-        return output
+            filtered_outputs = []
+            for output in outputs:
+                filtered_output = self.filter_output(output)
+                filtered_outputs.append(self.handle_output(mode, filtered_output, forced))
+            outputs = filtered_outputs
+        return '\n'.join(outputs)
 
     def handle_output(self, mode, content, forced, basename=None):
         # Then check for the appropriate output method and call it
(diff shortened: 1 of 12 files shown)


karyon:
please help me understand this. this PR

1. Adds OFFLINE_GROUP_FILES that enables/disables concatenation of files in offline mode (this is what you mean by grouping, right?)
    * It does not add this feature for "online" compression, right?
    * what is the use case for this setting? is not concatenating files a requirement for source maps?
1. Adds CLOSURE_COMPILER_SOURCEMAPS which makes the closure compiler generate source maps
1. Adds OFFLINE_SOURCEMAPS_ON_FILES
    * this i don't quite get. do source maps work without making this true? what's the benefit of encoding this in base64 instead of a normal .map file?

ciotto:
Hello,
I try to explain:

 1. I probably used the wrong name for variable, I did a check and if `OFFLINE_GROUP_FILES` is False, the concatenation of files is disabled also in "on-line" mode. The correct name can be `ENABLE_FILES_CONCATENATION`. I added this function becouse, for the flow of compressor, I can't minify more files in one Closur Compiler call, so I can't generate a single .map file for more files.

 2. Correct

 3. This appear to work also in "on-line" mode, so i used  the wrong name. I used the base64 min because the [SASS precompiler](https://github.com/torchbox/django-libsass) used this before and because I can't generate the file in the filter, but base64 source map is not largely supported and is not good for download size

Now I use this code in production an it work fine.

I will change the variables names as sun as possible.

karyon:
ref #438.

closing due to inactivity.

as far as i can see (correct me if i'm wrong), this PR makes the clojure compiler output sourcemaps into files, appends that as a base64 data url to the js file, and then decodes the base64 data, writes it to a file, and replaces the data url with a url pointing to that file. i think this is the wrong approach, a better one would make the compiler put the source map more directly into the storage.

ciotto:
I'm sorry for my inactivity, but, even though I use this code in production, I opted for a more usable approach, so I don't spent other time on this develop.

Surely my approach is not the best but, if I remember, *CosureCompiler* can't write source maps in a separated file. Anyhow, write directly on a storage require a big refactor of the Compressor's logics.

In other words, this is a workaround but work for my purpose.

karyon:
yeah, "require a big refactor" would also be my first guess. however, django-compressor is quite complex and hard to maintain as it is, and bolting on more workarounds will only worsen this :)

PR #817, opened 2017-01-14
Add support for PostCSS

Pithikos:
I thought it would be advantageous to be able and use PostCSS with the myriad of plugins it offers like the famous Autoprefixer. I did find [this implementation](https://github.com/dizballanze/django-compressor-autoprefixer) which however is specific to only one plugin.

With the implementation here you can use any number of plugins by specifying the `COMPRESS_POSTCSS_PLUGINS` variable in settings.py.

I did my best to add documentation and tests and follow the guidelines for `django-compressor`. I hope this gets merged since I think it can better the workflow for many people (it sure will for me!).

I tested only with Python2.7 and Python3.5 with no error messages. I also wanted to add tests that make usage of the plugins themselves but that would require using `npm` in the testsuite and since I couldn't find anything like that already there I decided to skip it.

Feel free to comment..

### DIFF
--- a/compressor/filters/postcss.py
+++ b/compressor/filters/postcss.py
@@ -1,0 +1,33 @@
+from compressor.conf import settings
+from compressor.filters import CompilerFilter
+
+
+DEFAULT_BINARY = "postcss"
+DEFAULT_ARGS = " "
+DEFAULT_PLUGINS = ()
+
+
+def fetch_options():
+    def plugins_as_args(plugins):
+        return ''.join(map(lambda plugin: '--use %s ' % plugin, plugins))
+    return (
+        ("binary", getattr(settings, "COMPRESS_POSTCSS_BINARY", DEFAULT_BINARY)),
+        ("args", getattr(settings, "COMPRESS_POSTCSS_ARGS", DEFAULT_ARGS)),
+        ("plugins", plugins_as_args(getattr(settings, "COMPRESS_POSTCSS_PLUGINS", DEFAULT_PLUGINS))),
+    )
+
+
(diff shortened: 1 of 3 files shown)


diox:
Hi,

Sorry to be negative, but there is little change for this to be merged.

I'd like to avoid shipping many filters with compressor and encourage people to build their own according to their needs - or just stop using compressor altogether when they start having complex requirements, replacing it with a more modern front-end stack.

Pithikos:
That is understandable. Probably it might be worth having something about this in the contribution section of the docs since it's unclear and only a reference to ask at IRC (which is not that active) is mentioned.

In any case I made this as a [separate project](https://github.com/Pithikos/django-compressor-postcss) so feel free to close the pull request.

Cheers

thibault:
I know this is an old issue but since I had the same problem and google sent me here for a solution, here's how I solved this with a minimal amount of overhead.

First, I installed `postcss` and `autoprefixer` at the root of my django project.

     npm install postcss-cli autoprefixer --save

In my django settings:

```python
NODE_MODULES_PATH = DJANGO_ROOT.child('node_modules')

SASS_COMPILER = 'make css'

COMPRESS_PRECOMPILERS = (
    ('text/x-scss', '{} include={} infile={{infile}} outfile={{outfile}}'.format(
        SASS_COMPILER, NODE_MODULES_PATH)),
)
```

In a `Makefile`:

```make
css:
	sass -I $(include) $(infile) $(outfile)
	./node_modules/.bin/postcss --use autoprefixer -r $(outfile)
```

And that's all folks.

PR #822, opened 2017-02-06
clarified documentation for COMPRESS_CACHE_BACKEND

Arany:
When using the "default" ``CACHES["default"]`` as a value for ``COMPRESS_CACHE_BACKEND`` an error would be thrown. The correct way is to use the name of the cache backend.


karyon:
The whole paragraph was severely outdated, i have updated it in #851. Thanks for bringing my attention to this!

PR #830, opened 2017-02-19
add ldjson to compress

darkkase:
i add ld-json for schema.org https://developers.google.com/schemas/formats/json-ld

before:
if you use application/ld+json  and put the code inside {%compress js inline%} {%endcompress%} django-compressor change the type "application/ld+json" to "text/javascript"

after:
now you can use  {%compress ldjson inline%} and...  {%compress ldjson %}


diox:
Sorry, I don't think this use case is common enough to warrant the inclusion of this fix in compressor core.

dkoldyaev:
@darkkase I use the small templatetag for compress json https://djangosnippets.org/snippets/10661/
maybe it'll help you

PR #835, opened 2017-03-13
Add test making sure {% compress <mode> inline %} works.

diox:


diox:
Ah yeah, I had not updated, this is because of the double `;;`, I'll update.

Issue #706, opened 2015-12-16
`ValueError` with 1.6 on some pages only.


mrmachine:
Here's the template code and traceback:

```
Template error:
In template /path/to/myproject/templates/myproject/base.html, error at line 84
   dictionary update sequence element #0 has length 22; 2 is required
   74 : 	{% endif %}
   75 : 
   76 : 	{% block icons %}
   77 : 		<link rel="shortcut icon" href="{% static "myproject/images/icons/favicon.ico" %}">
   78 : 		<link rel="apple-touch-icon" href="{% static "myproject/images/icons/apple-touch-icon.png" %}">
   79 : 		<link rel="apple-touch-icon-precomposed" href="{% static "myproject/images/icons/apple-touch-icon-precomposed.png" %}">
   80 : 	{% endblock icons %}
   81 : {% endblock %}
   82 : 
   83 : {% block base_css %}
   84 : 	 {% compress css file %} 
   85 : 		<link rel="stylesheet" href="{% static "bootstrap/dist/css/bootstrap.css" %}">
   86 : 		<link rel="stylesheet" href="{% static "font-awesome/css/font-awesome.css" %}">
   87 : 		<link rel="stylesheet" href="{% static "slick-carousel/slick/slick.css" %}">
   88 : 		<link type="text/less" rel="stylesheet" href="{% static "myproject/less/main.less" %}">
   89 : 	{% endcompress %}
   90 : {% endblock %}
   91 : 
   92 : {% block head_js %}
   93 : 	{% compress js file %}
   94 : 		<script src="{% static "modernizr/modernizr.js" %}"></script>

Traceback:
File "/path/to/venv/lib/python2.7/site-packages/django/core/handlers/base.py" in get_response
  111.                     response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "/path/to/venv/lib/python2.7/site-packages/django/db/transaction.py" in inner
  394.                 return func(*args, **kwargs)
File "/path/to/venv/src/django-multiurl/multiurl.py" in multiview
  72.                     response = match.func(request, *match.args, **match.kwargs)
File "/path/to/venv/src/django-multiurl/multiurl.py" in multiview
  72.                     response = match.func(request, *match.args, **match.kwargs)
File "/path/to/venv/lib/python2.7/site-packages/django/contrib/auth/decorators.py" in _wrapped_view
  21.                 return view_func(request, *args, **kwargs)
File "/path/to/myproject/research_projects/views.py" in document
[...]

PR #836, opened 2017-03-13
Seems to fix problem with context dictionary update sequence length

cisarpavel:
Fix for retreiving final context in base.py/render_output.
Usage of flatten func is not controlled by existence of corresponding function but by django version.
Issue https://github.com/django-compressor/django-compressor/issues/706


karyon:
this is likely the wrong solution, see #706 for an explanation.

PR #840, opened 2017-04-10
Check for a trailing slash according to the operative system

glaand:


karyon:
i just checked and URLs always use forward slashes. what was the problem you were solving?

PR #846, opened 2017-05-04
Allow unicode in error message

logston:


karyon:
as far as i understand, this should not change anything since we have `from __future__ import unicode_literals` in the file. additionally, i was not able to reproduce any error or broken characters with the current code. what was the problem you were trying to solve?

logston:
Hmmm. I don't have the exact error anymore but it contained the following string:

```
u"\u2018env\u2019"
```

And the Django command "compress" failed to correctly raise the `CommandError` error.

logston:
Happy to close this if you feel it is unnecessary. Just wanted to do my due diligence to give back.

karyon:
yeah and i'm grateful for that :) i just don't understand why your change fixes your problem, as we have that import there which should make all string literals unicode... if you could investigate, that would be nice, because i could not reproduce any such error. otherwise, i can do nothing more than close this...

logston:
Sorry, I don't have the time right now to replicate the error. Next time I'll save it!

PR #853, opened 2017-05-13
Fix #837 -- Make it much easier to debug `OfflineGenerationError`.

### LINKED ISSUE #837: Make `OfflineGenerationError` easier to debug
We have a base template with something like: `{% compress js %}{% block js %}{% endblock %}{% endcompress %}`. Any number of templates might extend this base template, and define their own `{% block js %}...{% endblock %}` content.

If any of those templates includes a template tag or variable that depends on context that is not available during offline generation, `OfflineGenerationError` exceptions will be raised.

The error page when this happens only highlights the `{% compress js %}` line of the base template, as a hint as to where the error might be occurring.

Then we basically need to search the entire project for every occurrence of `{% block js %}` and audit its content, looking for a dynamic variable or template tag.

Looking at the traceback, in `compressor.templatetags.compress.render_offline`, just before the `OfflineGenerationError` exception is raised, we first generate the key from the original content with `key = get_offline_hexdigest(self.get_original_content(context))`.

I think we should pass an extended message to `OfflineGenerationError` which includes not only the key, but also the original content that was used to generate the key. This will provide a **MUCH** stronger hint as to where we might find the offending dynamic content.

mrmachine:
Previously, we were only including the hex digest key derived from the
original content in the error message. This is useless because:

* The key is missing, so we can't search templates or the cache for it
  and find anything useful.

* We have no idea which template or compress block might have generated
  the key.

Including the original content gives us a very strong clue that will
help us identify the compress block which has some dynamic content that
is causing the error.

### DIFF
--- a/compressor/templatetags/compress.py
+++ b/compressor/templatetags/compress.py
@@ -61,7 +61,8 @@ class CompressorMixin(object):
         If enabled and in offline mode, and not forced check the offline cache
         and return the result if given
         """
-        key = get_offline_hexdigest(self.get_original_content(context))
+        original_content = self.get_original_content(context)
+        key = get_offline_hexdigest(original_content)
         offline_manifest = get_offline_manifest()
         if key in offline_manifest:
             return offline_manifest[key].replace(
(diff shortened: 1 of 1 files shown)


karyon:
nice, thanks!

PR #856, opened 2017-05-24
Python 3.6 invalid escape sequence deprecation fix

scop:
https://docs.python.org/3/whatsnew/3.6.html#deprecated-python-behavior


karyon:
hi, since that string is not a regex, i think the proper fix would be to use a double backslash.

what did you use to get that warning? if i run the tests with `python -Wall` i get some warnings but not this one.

scop:
"r" is not at all limited to regexps, it is just as proper a fix as double backslash would be. 

-Wall should give you the warning. Just make sure to use python 3.6.

karyon:
TIL that r means raw and not regex! thanks :)

no idea though why i don't get the warning. well, whatever...

Issue #775, opened 2016-06-22
Original filename lost after CSS filters are run


ukch:
Say we have a SASS file `base.scss`, which we include as follows (using `COMPRESS_PRECOMPILERS`):

``` django
{% compress css %}
<link rel="stylesheet" type="text/scss" href="{% static "/path/to/base.scss" %}">
{% endcompress %}
```

This gets turned into the following HTML:

``` html
<link rel="stylesheet" href="/static/CACHE/css/base.623d072c400f.css" type="text/css" />
```

(note the name `base.623d072c400f.css` still contains the original name 'base')

However, if we then define some filters in `COMPRESS_CSS_FILTERS` (for example, [django-compressor-autoprefixer](https://pypi.python.org/pypi/django-compressor-autoprefixer/0.0.1)), the generated HTML changes to this:

``` html
<link rel="stylesheet" href="/static/CACHE/css/048fe983d13b.css" type="text/css" />
```

(note that the name `048fe983d13b.css` no longer contains the word 'base')

Is there a way to retain the original filename in this situation?

karyon:
from a quick glance it seems like this behaviour depends on `COMPRESS_ENABLED` being True or not, not whether any filters are defined, although i'm not sure. do you by any chance remember whether you changed `COMPRESS_ENABLED` as well?

ukch:
@karyon sorry I don't remember, although that rings a bell and might certainly be the case. Either way, it would still be great to remember filenames in this case.

karyon:
i think this is mostly fixed or at least made obsolete by #861

PR #861, opened 2017-06-07
Ensure generated file always contains a base name

mgjbroadbent:
This changes the default output filename from [hash].[ext] to [name].[hash].[ext] in all cases.

This makes the output naming consistent with Whitenoise in all cases.  This is because only files named as <name>.<hash>.<ext> will be considered immutable, all other files will have the default 10 sec age set.

In addition, the user can customise the name using the block_name attribute.

* The default name is set to output.  
* Use the block_name as the name if it's provided in the template tag.
* When using single or multiple js/css files with no filters defined the basename of one of those files will still be used for the name (this is the existing behaviour).
* Tests added and updated to support the name change.

This change was originally to solve #775, but this only changes the output naming convention.

### DIFF
--- a/docs/usage.txt
+++ b/docs/usage.txt
@@ -27,7 +27,7 @@ Which would be rendered something like:
 
 .. code-block:: django
 
-    <link rel="stylesheet" href="/static/CACHE/css/f7c661b7a124.css" type="text/css" charset="utf-8">
+    <link rel="stylesheet" href="/static/CACHE/css/one.f7c661b7a124.css" type="text/css" charset="utf-8">
 
 or:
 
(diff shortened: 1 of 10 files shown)


docs/usage.txt
```diff
@@ -27,15 +27,15 @@ Which would be rendered something like:
 
 .. code-block:: django
 
-    <link rel="stylesheet" href="/static/CACHE/css/f7c661b7a124.css" type="text/css" charset="utf-8">
+    <link rel="stylesheet" href="/static/CACHE/css/one.f7c661b7a124.css" type="text/css" charset="utf-8">
```
karyon: this should be `output.` instead of `one.`, right?

karyon:
@mgjbroadbent thank you very much for your contribution!

PR #865, opened 2017-07-06
Allow md5 hash to be overridden

shaunbrady:
On FIPS[0] enabled systems unsafe cryptographic modules are disabled.
This commit leaves the default of 'md5', but allows one to override
which hashlib library is loaded.

A test for sha256 is included.

AppConf was not used, as it appears the get_hexdigest function is called
too early or in the wrong context to utilize AppConf.

[0] http://csrc.nist.gov/groups/STM/cmvp/documents/140-1/140val-all.htm

### DIFF
--- a/compressor/cache.py
+++ b/compressor/cache.py
@@ -18,7 +18,8 @@ _cachekey_func = None
 
 
 def get_hexdigest(plaintext, length=None):
-    digest = hashlib.md5(smart_bytes(plaintext)).hexdigest()
+    hash_name = getattr(settings, 'COMPRESS_CACHE_HASH', 'md5')
+    digest = getattr(hashlib, hash_name)(smart_bytes(plaintext)).hexdigest()
     if length:
         return digest[:length]
     return digest
--- a/compressor/tests/test_base.py
+++ b/compressor/tests/test_base.py
@@ -211,6 +211,12 @@ class CompressorTestCase(SimpleTestCase):
         out = '<script type="text/javascript" src="/static/CACHE/js/d728fc7f9301.js"></script>'
         self.assertEqual(out, self.js_node.output())
 
+    @override_settings(COMPRESS_CACHE_HASH='sha256')
+    def test_js_output_sha256(self):
+        """ Modify the CACHE_HASH to use sha256 instead of md5 """
+        out = '<script type="text/javascript" src="/static/CACHE/js/74e158ccb432.js"></script>'
+        self.assertEqual(out, self.js_node.output())
+
     def test_js_override_url(self):
         self.js_node.context.update({'url': 'This is not a url, just a text'})
         out = '<script type="text/javascript" src="/static/CACHE/js/d728fc7f9301.js"></script>'


diox:
Although disabling `md5` entirely seems weird to me (in compressor we are not using it for security purposes and don't really care about its weaknesses - we're even shortening the resulting hash...) I would not mind switching to a different hash function without providing a setting to change this. The hash function would just need to be reasonably fast.

shaunbrady:
@carltongibson I can certainly elaborate... I was flat wrong.  In my testing I forgot COMPRESS_.  So if we do have a setting, I think I can utilize AppConf.  My apologies, saw some straight Django settings overrides (STATIC_URL) and got snow blind.

@diox Yeah, it's rather aggressive (disabling functions all together); I think the argument is a higher level of trust.  I realize you guys (and many others) aren't using it for cryptographic purposes; I'd like to see all of the hash functions take f(THISISNTFORSECURITYIPROMISE=True) but it's not my world....

Do you have a hash function in mind?  I need to redo this PR either way to use AppConf (or not if you want to avoid another setting).  I selected sha256 because it was known to be covered by FIPS.  We could try sha1; my concern is they'll consider that not FIPS compliant in the future as well.  We could look to non-cryptographic functions, but I can't opine to them with out Googling.

Thanks!

diox:
I don't have good hash suggestions, no. I'm open to anything :) We don't care about backwards-compatibility (changing the files hash when upgrading is fine), we just need something that has an implementation in the standard python library and is fast enough.

shaunbrady:
So sha256 is doubleish the speed of CRC32, but we're still under a second for 1400 or so files (everything in static that's a file):

```
$ python test.py
starting crc32 at 1499452457.41
ending crc32 at 1499452457.44 did 1415
starting sha256 at 1499452457.44
ending sha256 at 1499452457.55 did 1415
starting md5 at 1499452457.55
ending md5 at 1499452457.59 did 1415
```
(This is (fs) cache hot btw).  I'd propose going with sha256, reasons:

1) This will get your platform agnostics (don't want your tests to random fail on "this computer here")
2) The bottle neck (for us at least) is pushing out static files up to a hosting provider.  We compress once.  Maybe that's not always the case.
3) (The most relevant I see) CRC32 can have collisions.  Any hash function can, but CRC32 is much smaller than md5.  If you have a false cache hit, I believe the wrong file is served, right?  You're not going to see that with sha256 (else we have larger problems ;))

I'll code a PR with what ever is proposed, but this is my suggestion.

Thanks!

SB

shaunbrady:
Better wording (above) .....  sha256 is halfish the speed.

shaunbrady:
I was going to code up both CRC32 and sha256, submit them both and let you decide which to merge.  However the tests will need to have their hashes updated, so while tractable, it's a bit of tedium, so I want to only pass through that once.  Let me know which way to proceed.  Thanks!

karyon:
i would be fine with sha256, it seems highly unlikely we are bottlenecked by hashing speed anywhere where it matters. @diox?

btw `crc32(data) & 0xffffffff` should be consistent across platforms according tho [the docs](https://docs.python.org/3/library/zlib.html#zlib.crc32)

diox:
Let's go with `sha256` then.

shaunbrady:
Excellent!  I'll get a new PR up that just uses sha256 and updates the tests.  Is there any magic on what was hashed for the test input?  (I haven't cruised the tests other  than to find a spot for the new one).

shaunbrady:
Closing in favor of #866

PR #866, opened 2017-07-18
Md5 to sha256

shaunbrady:
On FIPS[0] enabled systems unsafe cryptographic modules are disabled.
This commit changes out the md5 function for sha256, both of hashlib.

[0] http://csrc.nist.gov/groups/STM/cmvp/documents/140-1/140val-all.htm


karyon:
thank you very much! lgtm, but i'll leave the decision to @diox

diox:
Thanks!

Issue #867, opened 2017-08-02
Add support brotli


lcd1232:
Simillar like `compressor.storage.GzipCompressorFileStorage` add brotli compression. On python you can use [brotlipy](https://pypi.python.org/pypi/brotlipy/0.7.0) library.

karyon:
closing, the PR is enough :)

PR #868, opened 2017-08-10
Brotli Storage

lcd1232:
Implements this - https://github.com/django-compressor/django-compressor/issues/867
- Read file by chunks for saving memory

### DIFF
--- a/compressor/storage.py
+++ b/compressor/storage.py
@@ -99,7 +99,7 @@ class BrotliCompressorFileStorage(CompressorFileStorage):
         compressed_path = '%s.br' % orig_path
 
         with open(orig_path, 'rb') as f_in, open(compressed_path, 'wb') as f_out:
-            compressor = brotlipy.Compressor()
+            compressor = brotli.Compressor()
             while True:
                 f_in_data = f_in.read(self.chunk_size)
                 compressed_data = compressor.compress(f_in_data)
--- a/compressor/storage.py
+++ b/compressor/storage.py
@@ -92,9 +92,13 @@ class BrotliCompressorFileStorage(CompressorFileStorage):
     """
     chunk_size = 1024
 
-    def get_compressor(self):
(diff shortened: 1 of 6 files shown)


compressor/storage.py
```diff
@@ -85,6 +86,37 @@ def save(self, filename, content):
         return filename
 
 
+class BrotliCompressorFileStorage(CompressorFileStorage):
+    """
+    The standard compressor file system storage that gzips storage files
+    additionally to the usual files.
```
karyon: this needs updating

compressor/storage.py
```diff
@@ -85,6 +86,37 @@ def save(self, filename, content):
...
+    """
+    chunk_size = 1024
```
karyon: is this really necessary? i mean, we are talking about a few megabytes of data at the maximum, aren't we? were you running into any memory usage issues?

compressor/storage.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import unicode_literals
 import errno
 import gzip
+import brotli
```
karyon: this imports unconditionally, but it is not listed in the requirements in setup.py. Since most other dependencies we have are optional, i suggest to do the same here, an example is here: https://github.com/django-compressor/django-compressor/blob/develop/compressor/parser/beautifulsoup.py#L13

compressor/storage.py
lcd1232: Maybe someone will be using it for compressing video files or something big. Anyway you do not lose anything using this code.

lcd1232:
@karyon I add documentation and fix all problems.

compressor/storage.py
lcd1232: Fixed

compressor/storage.py
lcd1232: Fixed

karyon:
thank you very much!

Issue #870, opened 2017-08-16
Optimize compress offline loop(s)


Evelf:
Multiple imbricated loops are needs to generate the compressed files and manifest, but the way it's currently done is wasteful in memory when we have multiple contexts.

With a few meticulous changes, it can be drastically improved, in my benchmarks it goes from \~2942Mb to \~174Mb in memory consumption with no impact on the execution time (\~ 3'15"). See the attached mprofile files.

[mprofile.zip](https://github.com/django-compressor/django-compressor/files/1228841/mprofile.zip)

karyon:
closing, the PR is enough :)

PR #871, opened 2017-08-16
ref #870 - improves memory management

Evelf:
Multiple imbricated loops are needs to generate the compressed files and manifest, but the way it's currently done is wasteful in memory when we have multiple contexts.

With a few meticulous changes, it can be drastically improved, in my benchmarks it goes from \~2942Mb to \~174Mb in memory consumption with no impact on the execution time (\~ 3'15"). See the attached mprofile files.

[mprofile.zip](https://github.com/django-compressor/django-compressor/files/1228841/mprofile.zip)

### DIFF
--- a/compressor/management/commands/compress.py
+++ b/compressor/management/commands/compress.py
@@ -168,13 +168,18 @@ class Command(BaseCommand):
                                   (settings.COMPRESS_OFFLINE_CONTEXT, e))
         elif not isinstance(contexts, (list, tuple)):
             contexts = [contexts]
-        contexts = list(contexts) # evaluate generator
 
         parser = self.__get_parser(engine)
-        compressor_nodes = OrderedDict()
+        fine_templates = []
+
+        log.write("Compressing... ")
+        verbosity = 1
+
         for template_name in templates:
             try:
                 template = parser.parse(template_name)
+                template.template_name = template_name
+                fine_templates.append(template)
             except IOError:  # unreadable file -> ignore
                 if verbosity > 0:
                     log.write("Unreadable template at: %s\n" % template_name)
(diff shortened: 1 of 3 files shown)


compressor/management/commands/compress.py
```diff
@@ -287,6 +285,9 @@ def handle_extensions(self, extensions=('html',)):
         return set(ext_list)
 
     def handle(self, **options):
+        self.handle_inner(**options)
+
+    def handle_inner(self, **options):
```
karyon: in how far does this help?

compressor/offline/jinja2.py
```diff
@@ -84,7 +84,7 @@ def get_init_context(self, offline_context):
 
         # Allow offline context to override the globals.
         context = self.env.globals.copy()
-        context.update(offline_context)
+        context.update(flatten_context(offline_context))
```
karyon: could you explain this change?

compressor/tests/test_offline.py
```diff
@@ -560,21 +562,10 @@ class OfflineCompressTestCaseErrors(OfflineTestCaseMixin, TestCase):
...
-            # Because we use env.parse in Jinja2Parser, the engine does not
-            # actually load the 'extends' and 'includes' templates, and so
-            # it is unable to detect that they are missing. So all the
-            # 'compress' nodes are processed correctly.
-            self.assertEqual(4, count)
-            self.assertEqual(engine, 'jinja2')
-            self.assertIn(self._render_link('7ff52cb38987'), result)
-            self.assertIn(self._render_link('2db2b4d36380'), result)
-
+        self.assertEqual(2, count)
```
karyon: could you explain this?

compressor/management/commands/compress.py
Evelf: We need ```handle_inner()``` return values for the tests, but Django's ```handle()``` can not return anything.

compressor/tests/test_offline.py
Evelf: Seems that this PR "fixes" a bug in compressor management of jira's templates, there's only 2 non bugged templates and compressor should return only 2 results for both engines.

But I wanted to look a bit more closely at this, and forgot to do so before pushing the PR, thanks for pointing this again to me :)

compressor/offline/jinja2.py
Evelf: It's a relic of the first attempt to improve memory consumption, it's needed to be able to use a generator for the context function.
It may or may not be part of this PR, as you prefer. In the end, that's not how performance where improved, but it's nice that you can give a generator to compressor without errors.

karyon:
continued in #884, thanks :)

compressor/tests/test_offline.py
karyon: this happened due to an introduced bug: the call to `compress` that got changed to `handle` required changing the parameter `engine` to `engines`, otherwise the parameter `engine` was not set, defaulted to `django`, and `jinja2` wasn't tested at all.

PR #876, opened 2017-08-28
Offline de-compression for string-alike STATIC_URL and COMPRESS_URL

kottenator:
Django doesn't incorporate `SCRIPT_NAME` WSGI param into `settings.STATIC_URL` so I've created `ScriptNamePrefixed(str)` subclass which dynamically adds `SCRIPT_NAME` prefix to itself:

```py
# settings.py
STATIC_URL = ScriptNamePrefixed('/static/')
```

```py
from django.conf import settings
settings.STATIC_URL == '/static/'
# Now WSGI param SCRIPT_NAME is /some/prefix/
settings.STATIC_URL == '/some/prefix/static/'
```

`ScriptNamePrefixed` re-defines all the `str` methods, introducing _the prefix_. However, I can't trick CPython's `str.replace(ScriptNamePrefixed('/static/'), '')` - it ignores _the prefix_ and uses the original string value. But if I replace: 

```py
# compressor/cache.py:55
render_template_string.replace(settings.STATIC_URL, '')
# with
render_template_string.replace(str(settings.STATIC_URL), '')
```
... and:

```py
# compressor/templatetags/compress.py:68
offline_manifest[key].replace(settings.COMPRESS_URL_PLACEHOLDER, settings.COMPRESS_URL)
# with
offline_manifest[key].replace(settings.COMPRESS_URL_PLACEHOLDER, str(settings.COMPRESS_URL))
``` 
&ndash; it works.

This PR adds ability for `settings.STATIC_URL` and `settings.COMPRESS_URL` to be anything that re-defines `__str__` method, not only strings.

### DIFF
--- a/compressor/base.py
+++ b/compressor/base.py
@@ -5,6 +5,7 @@ from importlib import import_module
 
 from django import VERSION
 from django.core.files.base import ContentFile
+from django.utils import six
 from django.utils.safestring import mark_safe
 from django.utils.six.moves.urllib.request import url2pathname
 from django.template.loader import render_to_string
@@ -79,6 +80,13 @@ class Compressor(object):
             base_url = self.storage.base_url
         except AttributeError:
             base_url = settings.COMPRESS_URL
+
+        # Cast ``base_url`` to a string to allow it to be
+        # a string-alike object to e.g. add ``SCRIPT_NAME``
+        # WSGI param as a *path prefix* to the output URL.
+        # See https://code.djangoproject.com/ticket/25598.
+        base_url = six.text_type(base_url)
+
         if not url.startswith(base_url):
             raise UncompressableFileError("'%s' isn't accessible via "
                                           "COMPRESS_URL ('%s') and can't be "
--- a/compressor/cache.py
+++ b/compressor/cache.py
@@ -7,6 +7,7 @@ from importlib import import_module
 
 from django.core.cache import caches
 from django.core.files.base import ContentFile
+from django.utils import six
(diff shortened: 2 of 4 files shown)


diox:
It would be best to use `six.text_type` (from `django.utils` to stay consistent with the rest of the code) for this instead of `str`

karyon:
i have zero idea what this does and it seems to be quite an edge case, but if @diox says it's okay then it is okay. but i'd suggest adding at least a comment above the two lines, if not tests, to explain why we should cast a setting that's usually a string to a string.

kottenator:
@karyon, @diox - does it look good now?

karyon:
fine by me, but i'll leave merging to @diox as i'm unsure what this does.

diox:
This still needs a test (especially as it's a pretty obscure feature that we're likely to break without realizing)

kottenator:
Sure, I'll add it.

karyon:
ping :)

kottenator:
@karyon - pong :) 
Finally, I have some time to work on my PR. Will add tests today.

kottenator:
@karyon, @diox - I've added tests. I've also added some explanation of my motivation. I hope you'll find it useful.

Please, let me know what else can/should I do here ;)

diox:
Looks good. The repetition of the comment is a bit un-DRY, but that's fine, we can live with that.

diox:
Thank you for your patience :) This is now merged.

PR #884, opened 2017-10-08
ref #870 - improves memory management

karyon:
this is #871 with some fixes in the second commit as well as some more cleanups.


karyon:
Thanks to @Evelf again!

Evelf:
Thanks to you @karyon ! :sparkles:

PR #682, opened 2015-10-22
Adding sekizai post-processor specialized for css and js independently of sekizai template tag

Alcolo47:
Allowing writing:
{% render_block "myJsBlock" postprocessor "compressor.contrib.sekizai.compressJs" %}
{% render_block "myCssBlock" postprocessor "compressor.contrib.sekizai.compressCss" %}

PR #891, opened 2017-12-04
Enhance Sekizai compressor support

jrief:
**django-compressor**'s Sekizai plugin, currently is not able to handle files from external sources such as CDNs. Therefore one would have to use another prefix, say ``{% addtoblock "ext-css" %}`` and ``{% addtoblock "ext-js" %}`` when adding these references. This however causes inconsistencies and other issues.

I therefore enhanced the Sekizai processor, so that it only concatenates files available locally. External sources are just served using a reference.

### DIFF
--- a/compressor/contrib/sekizai.py
+++ b/compressor/contrib/sekizai.py
@@ -22,14 +22,14 @@ def compress(context, data, name):
     """
     # separate compressable from uncompressable files
     parser = get_class(settings.COMPRESS_PARSER)(data)
-    compressor = Compressor()
+    js_compressor, css_compressor = Compressor('js'), Compressor('css')
     compressable_elements, expanded_elements, deferred_elements = [], [], []
     if name == 'js':
         for elem in parser.js_elems():
             attribs = parser.elem_attribs(elem)
             try:
                 if 'src' in attribs:
-                    compressor.get_basename(attribs['src'])
+                    js_compressor.get_basename(attribs['src'])
             except UncompressableFileError:
                 if 'defer' in attribs:
                     deferred_elements.append(elem)
@@ -42,7 +42,7 @@ def compress(context, data, name):
             attribs = parser.elem_attribs(elem)
             try:
                 if parser.elem_name(elem) == 'link' and attribs['rel'].lower() == 'stylesheet':
-                    compressor.get_basename(attribs['href'])
+                    css_compressor.get_basename(attribs['href'])
             except UncompressableFileError:
                 expanded_elements.append(elem)
             else:
--- a/compressor/contrib/sekizai.py
+++ b/compressor/contrib/sekizai.py
@@ -6,6 +6,11 @@
  and: https://github.com/ojii/django-sekizai.git@0.6 or later
 """
 from compressor.templatetags.compress import CompressorNode
+from compressor.exceptions import UncompressableFileError
+from compressor.base import Compressor
+from compressor.conf import settings
+from compressor.utils import get_class
+
 from django.template.base import TextNode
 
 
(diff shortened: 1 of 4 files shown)


jrief:
I know, tests are missing, but they were missing already in the first place. Have to see, how to add them.

karyon:
i don't quite understand, why is this problem specific to the sekizai extension? i mean either django-compressor does not handle that at all, then it wouldn't make sense to fix this just for the sekizai extension, or django-compressor handles this, then i wonder why so much additional code is necessary. can you help me understand that? also, can you explain in how far this supersedes #682?

and yes, this needs tests :) you could check out test_jinja2ext, which tests the other file in the contrib folder.

jrief:
The intention of Sekizai is to bundle JavaScript and Stylesheets together with their HTML components. This means that inside those HTML snippets one would add ``{% addtoblock "css" %}...{% endaddtoblock  %}`` and ``{% addtoblock "js" %}...{% endaddtoblock  %}``.

Sometimes however, one can not refer to local files but must rather include them from a CDN. Here the Sekizai compressor must distinguish between local files and external. One way I did this in the past was to use ``{% addtoblock "ext-js" %}...{% endaddtoblock  %}`` and then render that block without using the compressor.

This however is an unportable solution, because an app using ``{% addtoblock "ext-js" %}`` doesn't know if the projects renders these imports in the correct order.

About testing, I already started to implement them using the Jinja2 extension test as blueprint.

karyon:
in how far is the ordering issue fixed, since you are re-ordering the js files as well?

also, can you check what django-compressor does with external and deferred files in one compress block when not using sekizai?

jrief:
currently writing unit tests is stalled, because of other issues.

karyon:
hi! is there any chance you continue working on this? to me it would be important to clarify what django-compressors behaviour in such cases without sekizai support looks like. i'm willing to look into this myself, but only if there's a chance to get this finished :)

jrief:
Sorry, I was distracted from writing the unit tests.

Just for your information, the postprocessor works fine and I use it all the time in my own projects. The only missing part is the unit test. Hope to find some time in the next months.

jrief:
Unit tests finally have been added. So this feature could be released with the next version.

I also added some docs which explains why this special postprocessor is required in combination with Sekizai.

jrief:
@karyon Unit tests have been added more than a month ago. From your point of view, is there still anything missing?

karyon:
sorry.

in how far can this create new ordering issues, since you are effectively re-ordering js files? even though it might be a consistent and predictable re-ordering but still, that might swap the ordering of files that depend on one another, no? in how far can this break existing projects that update to the new version and get their js files reordered?

also, could you shed some light on why our sekizai plugin should have that feature of filtering out external js files while vanilla django-compressor should not do that?

jrief:
@karyon in existing projects, mixing external with internal files was impossible anyway, so there can't be any re-ordering issue.

> also, could you shed some light on why our sekizai plugin should have that feature of filtering out external js files while vanilla django-compressor should not do that?

Sekizai's big plus is that you can perform a good separation of concern for JS and CSS files. This means that every template snippet only has to add the assets it requires for itself. Say you have a Django template snippet, then you would add this to it somewhere.

```
(...)
{% addtoblock "css" %}<link href="{% static 'myapp/css/file.css' %}" rel="stylesheet" />{% endaddtoblock %}
{% addtoblock "js" %}<script src="{% static 'myapp/js/file.js' %}"></script>{% endaddtoblock %}
(...)
```

If a CSS or JS file must be referenced through a CDN rather than though the local file system, then this approach works as well – until you use django-comporessor.
This is because django-comporessor can not – for obvious reasons – concatenate and compress files located outside its realm. Therefore, this statement
```
{% render_block "<css/js>" postprocessor "compressor.contrib.sekizai.compress" %}
```
would fail with an `UncompressableFileError `.

A workaround would be to use another namespace for your Sekizai blocks, for instance
```
(...)
{% addtoblock "ext-css" %}<link href="https://code.jquery.com/ui/1.12.1/themes/base/jquery-ui.css" rel="stylesheet" />{% endaddtoblock %}
{% addtoblock "ext-js" %}<script src="https://code.jquery.com/jquery-3.3.1.min.js"></script>{% endaddtoblock %}
(...)
```

and add two additional statements to the project's main template, such as
```
{% render_block "<ext-css/ext-js>"  %}
```
this time without a postprocessor.

This however does not work well across apps boundaries, because the convention requires to use 
`{% addtoblock "css" %}...` and `{% addtoblock "js" %}...` respectively. In Sekizai there is no convention to use `ext-css` and `ext-js` for files served from a CDN, because Sekizai doesn't care about compression, so without django-compressor it's not a problem for Sekizai.

It also becomes a problem for projects starting with Sekizai alone, using ``{% addtoblock "css" %}...`` and ``{% addtoblock "js" %}...`` for both, local static files and files served through a CDN. Then, when they later want to add django-compressor together with Sekizai's postprocessor, they have to rewrite all their templates, separating the ``addtoblock``s into `ext-css` and `ext-js`. This is quite annoying and causes a lot of trouble for newbies.  

Since I often ran into this problem, I wrote this enhanced version of a Sekizai postprocessor. Currently I add it to all of my projects as a stand-alone module. Up to now I never had any ordering issues or any other side-effect.

karyon:
thank you. i don't have the energy to follow your explanation in detail, but i'll merge anyways.

PR #893, opened 2017-12-14
Added mako template support in contrib

WnP:
## Usage

All you need to do is to import extension namespace and using it with mako syntax:

```mako
<%namespace name="compressor" module="compressor.contrib.mako"/>

<%compressor:compress kind="css">
<link rel="stylesheet" href="/static/css/style.css" type="text/css" />
</%compressor:compress>
```

All django-compressor arguments are supported but must be explicitly named:

* ``kind``: mandatory, "css" or "js".
* ``mode``: optional, compressor mode, "file" or "inline".
* ``name``: optional but requires ``mode``, output file name.

### DIFF
--- a/compressor/contrib/mako.py
+++ b/compressor/contrib/mako.py
@@ -1,0 +1,68 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+from mako.runtime import supports_caller
+from compressor.templatetags.compress import compress as compressor
+
+
+class NodeMock(object):
+    def __init__(self, content, *args, **kwargs):
+        self.content = content
+
+    def render(self, *args, **kwargs):
+        return self.content
+
+
+class ParserMock(object):
+    def __init__(self, body, *args, **kwargs):
+        self.body = body
+
+    def parse(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


cdevienne:
How about exporting a "js" and  a "css" function ?

We would have:

```mako
<%namespace name="compress" module="compressor.contrib.mako"/>
<%compress:css>
    <link rel="stylesheet" href="/static/css/style.css" type="text/css" />
<%/compress:css>
```

Aside from being a little cleaner, it would make it impossible to forget the 'kind' argument.

diox:
Hi, 

I'm not sure this should be in django-compressor itself. Can't you release it separately?

WnP:
@cdevienne sounds a good idea, by the way I'll preserve the original function too to preserve the same API

@diox like you want, I thought that like `jinja2` support is provided in `contrib` we can do the same for `mako`

diox:
Well jinja2 is a lot more popular than mako :) And we're trying to keep compressor lean so it's not a nightmare to maintain.

I'd accept a PR to add a mention to an external package adding support for mako in the docs.

PR #894, opened 2017-12-19
adding a setting to choose the hashlib

matthare:
Here at ORM, we have a custom compressor filter to find the names of files created when collectstatic is run. Our filter uses the get_hashed_content function, which consequently broke when 
Django-compressor changed from md5, which Django's ManifestStaticFilesStorage uses, to sha256.
We would love if Django-compressor allowed the user to choose which hashlib function to use.

### DIFF
--- a/compressor/cache.py
+++ b/compressor/cache.py
@@ -19,7 +19,9 @@ _cachekey_func = None
 
 
 def get_hexdigest(plaintext, length=None):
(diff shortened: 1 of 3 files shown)


karyon:
if i understand correctly, this only removes the need for the one-time transition to the new hash algorithm right? if yes, i would rather not merge this, since your case seems to be a rather special one...

karyon:
closing. thank you nonetheless!

PR #605, opened 2015-03-20
Fix crash in offline compression when child nodelist is None

aumo:
If a Node had an attribute set to `None` present in `child_nodelist`, `DjangoParser.get_nodelist` would fail.
It happened to me while using compressor in conjunction with sorl-thumbnail.
I have not enough knowledge of the Django template engine to know if sorl-thumbnail is doing something wrong or not but anyway it does not hurt to be more robust.

PR #895, opened 2018-01-25
Fix bug in offline compression for Django template engine

jaddison:
If a Node had an attribute set to None present in child_nodelist, nodelist collection would fail. The change here is simply coding a bit more defensively.

This fixes #605 (which is closed, but not fixed). Related to https://github.com/jazzband/sorl-thumbnail/issues/470.

### DIFF
--- a/compressor/offline/django.py
+++ b/compressor/offline/django.py
@@ -132,7 +132,9 @@ class DjangoParser(object):
         nodelist = []
         if isinstance(node, Node):
             for attr in node.child_nodelists:
-                nodelist += getattr(node, attr, [])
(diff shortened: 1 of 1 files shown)


diox:
Hi,

Can you provide a test for this change?

karyon:
we decided to get over it so here we go

vitalik:
any idea when will this land to pip ?

vitalik:
maybe this should be cherrypicked to version and uploaded to pypi ?

karyon:
well, to tell the truth, we probably won't do that. a proper release will be done at some point, and until then using the development version is really easy with pip.

PR #899, opened 2018-03-03
Updated jinja2 extension to behave similar to the django tag

WoLpH:
There were several inconsistencies between the Django and the Jinja2 tags which made the docs unclear:

1. The jinja2 command required a comma between the kind and mode arguments. This change omits the need (but still supports) the comma.
2. The previous command didn't support a filename, this adds optional support for that

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -22,9 +22,9 @@ Atamert Ölçgen
 Aymeric Augustin
 Bartek Ciszkowski
 Ben Firshman
-Ben Spaulding
 Benjamin Gilbert
 Benjamin Wohlwend
+Ben Spaulding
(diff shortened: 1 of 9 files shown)


diox:
Hi,

This is cool, but it needs tests to be merged.

WoLpH:
That's fair :)
I suppose I can simply copy the Django template tests now, they should work the same

WoLpH:
I've added a bunch of tests and the only thing I found that isn't working (and not easily fixable, I've tried) is the `super()` call with offline compression. But that's something that never worked before anyhow :)

karyon:
nice, thank you!

PR #900, opened 2018-03-08
Abstract away resource types to allow the implementation of new types.

alvra:
This change makes it possible to compress other resources besides css and js, such as svg, jpeg, fonts etc.

It introduces a new setting for filters ``COMPRESS_FILTERS`` that combines the existing ``COMPRESS_CSS_FILTERS`` and ``COMPRESS_JS_FILTERS``. The existing settings are still supported, but can optionally be replaced by the new one. This change is fully backwards-compatible and tested.

The undocumented setting ``CSS/JS_COMPRESSORS`` are combined into ``COMPRESSORS``. This is not backwards-compatible nor tested.

You can now add your own compressors.
In the example below, compressor now knows the ``jpeg`` resource type. Of course, the third party jpeg compressor still has to implement it.
```
# settings.py
COMPRESSORS = dict(
    css='compressor.css.CssCompressor',
    js='compressor.js.JsCompressor',
    jpeg='third.party.jpeg.JpegCompressor',
    ...
)
FILTERS = dict(
    jpeg=['third.party.jpeg.JpegOptimizerFilter'],
)

# page.html
<body>
  {% compress jpeg %}
    <img src="/path/to/image.jpeg" />
  {% endcompress %}
</body>
```

### DIFF
--- a/compressor/base.py
+++ b/compressor/base.py
@@ -31,22 +31,24 @@ class Compressor(object):
     depending implementations details.
     """
 
-    output_prefix = 'compressed'
     output_mimetypes = {}
 
     def __init__(self, resource_kind, content=None, output_prefix=None,
                  context=None, filters=None, *args, **kwargs):
         if filters is None:
-            filters = settings.COMPRESS_FILTERS[resource_kind]
-        self.content = content or ""  # rendered contents of {% compress %} tag
-        if output_prefix:
+            self.filters = settings.COMPRESS_FILTERS[resource_kind]
+        else:
+            self.filters = filters
+        if output_prefix is None:
+            self.output_prefix = resource_kind
+        else:
             self.output_prefix = output_prefix
+        self.content = content or ""  # rendered contents of {% compress %} tag
         self.output_dir = settings.COMPRESS_OUTPUT_DIR.strip('/')
         self.charset = settings.DEFAULT_CHARSET
         self.split_content = []
         self.context = context or {}
         self.resource_kind = resource_kind
-        self.filters = filters
         self.extra_context = {}
         self.precompiler_mimetypes = dict(settings.COMPRESS_PRECOMPILERS)
         self.finders = staticfiles.finders
@@ -229,10 +231,6 @@ class Compressor(object):
             if enabled:
                 yield self.filter(value, self.cached_filters, **options)
             elif precompiled:
-                # Since precompiling moves files around, it breaks url()
-                # statements in css files. therefore we run
-                # the absolute and relative filter on precompiled css files
-                # even if compression is disabled.
                 for filter_cls in self.cached_filters:
                     if filter_cls.run_with_compression_disabled:
                         value = self.filter(value, [filter_cls], **options)
--- a/compressor/base.py
+++ b/compressor/base.py
@@ -16,7 +16,6 @@ from compressor.conf import settings
 from compressor.exceptions import (CompressorError, UncompressableFileError,
         FilterDoesNotExist)
 from compressor.filters import CachedCompilerFilter
-from compressor.filters.css_default import CssAbsoluteFilter, CssRelativeFilter
 from compressor.storage import compressor_file_storage
 from compressor.signals import post_compress
 from compressor.utils import get_class, get_mod_func, staticfiles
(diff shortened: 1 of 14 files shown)


compressor/conf.py
```diff
@@ -131,3 +141,30 @@ def configure_precompilers(self, value):
...
+                            main_setting=self._meta.prefixed_name('FILTERS'),
+                            kind=kind))
+                data['FILTERS'][kind] = filters
+            elif kind in data['FILTERS']:
+                # filters for this kind are set using FILTERS[<kind>]
+                data[setting_name] = data['FILTERS'][kind]
+            else:
+                # filters are not defined
+                data['FILTERS'][kind] = data[setting_name] = default_filters[kind]
+        return data
```
karyon: i would rather settle on one format and let this code here transfer the other format if necessary, instead of ensuring both formats are present. what do you think?

compressor/filters/base.py
```diff
@@ -40,6 +40,13 @@ class FilterBase(object):
     Subclasses should implement `input` and/or `output` methods which must
     return a string (unicode under python 2) or raise a NotImplementedError.
     """
+
+    # Since precompiling moves files around, it breaks url()
+    # statements in css files. therefore we run the absolute and relative filter
+    # on precompiled css files even if compression is disabled.
+    # This flag allows those filters to do so.
+    run_with_compression_disabled = False
```
karyon: i would keep the comment only here and remove it everywhere else.

compressor/base.py
```diff
@@ -32,21 +31,36 @@ class Compressor(object):
     depending implementations details.
     """
 
-    def __init__(self, content=None, output_prefix=None,
+    output_prefix = 'compressed'
+    output_mimetypes = {}
+
+    def __init__(self, resource_kind, content=None, output_prefix=None,
```
karyon: what's the difference between resource_kind and output_prefix in practice? specifically, I am wondering whether there is any good reason to create e.g. a JsCompressor with a first arguent different from 'js'?

karyon: ...i now understand better that output_prefix is meant to control the output directory, which is now `compressed` by default for all filetypes other than js and css. how about making resource_kind another class variable?

compressor/base.py
```diff
@@ -260,7 +274,7 @@ def precompile(self, content, kind=None, elem=None, filename=None,
 
         filter_or_command = self.precompiler_mimetypes.get(mimetype)
         if filter_or_command is None:
-            if mimetype in ("text/css", "text/javascript"):
+            if mimetype in self.output_mimetypes:
```
karyon: ...funny, almost looks like it did allow css files in an `compress js` block (and vice versa)...

compressor/base.py
```diff
@@ -32,21 +31,36 @@ class Compressor(object):
...
+        if output_prefix:
+            self.output_prefix = output_prefix
         self.output_dir = settings.COMPRESS_OUTPUT_DIR.strip('/')
         self.charset = settings.DEFAULT_CHARSET
         self.split_content = []
         self.context = context or {}
-        self.type = output_prefix or ""
-        self.filters = filters or []
+        self.resource_kind = resource_kind
+        self.filters = filters
```
karyon: `self.filters = filters or settings.COMPRESS_FILTERS[resource_kind]`?

karyon:
also, please promise to write some documentation as soon as the code is settled :)

compressor/base.py
alvra: Good catch on `output_prefix` generally being the same as `resource_kind`,
I hadn't noticed about that.

I opted for making `resource_kind` an argument since it's
value is essentially controlled by the user in their settings.py.
Hardcoding it here would couple the kind to the compressor type,
which would probably result in breaking things.

It would also allow for running multiple version of one compressor
with different configurations, or using eg. a XmlCompressor
to compress generic xml, as well as html and svg.
Although I'm not sure how useful that would be in practice.

Since `output_prefix` is only used to vary the output path,
we could make it default to `resource_kind`
and remove the default class attribute.

compressor/base.py
alvra: That would remove the option of removing filters altogether. I will put an `if None else ` in between.

compressor/base.py
alvra: Yeah, I assumed that was an oversight and changed it to something simpler.

compressor/filters/base.py
alvra: Agreed

compressor/conf.py
alvra: I was going for something that supports both formats
during the transition, mainly since it makes the
code and testing a bit nicer IMO.

alvra:
Added a few changed as discussed.
I got started on docs for the new `COMPRESSOR_FILTERS` setting, but I prefer to figure out backwards-compatibility first.

compressor/base.py
karyon: makes sense and agreed.

compressor/conf.py
karyon: i don't quite understand, if the code outside of conf.py accessed only one of the two options (which it should in my opinion), then of what use is it to ensure the other one is present as well?

as to backwards-compatibility, the idea is of course to be backwards-compatible. 

i feel like we are misunderstanding each other, here an attempt to clarify my idea: just like it does right now in your PR, conf.py should be able to read both formats (while ensuring that only one is actually present to avoid conflicts), but in contrast to your PR, it should "output" only one format. the code outside of conf.py would read only that format. in my understanding, it should be strictly less code than the current PR.

alvra: I implemented the change as you suggested. The old form is no longer present after configuring.
Thanks for clarifying.

alvra:
First attempt at updating the docs to reflect the new settings, let me know what you think.

alvra:
Thanks! I've updated that last mention, let me know if you need anything else

karyon:
thank you very much!

PR #912, opened 2018-06-11
THREADING: Make template compression threaded to improve performance

keithhackbarth:
This is similar to the work done in https://github.com/django-compressor/django-compressor/pull/737. But since the task is I/O bound in most cases, it seemed to make more sense to choose threads over processes.

On our rather large site, using a custom UglifyJS filter and CSSMin, command line compression when from an average 12.38 minutes to an average 6.59 minutes

Note: initially I used locking to prevent a potential race-condition where the same hash could get compressed twice, but in my very non-scientific tests, performance was better without it.

### DIFF
--- a/compressor/management/commands/compress.py
+++ b/compressor/management/commands/compress.py
@@ -2,7 +2,7 @@ from __future__ import unicode_literals
 # flake8: noqa
 import os
 import sys
-import threading
+import concurrent.futures
 
 from collections import OrderedDict, defaultdict
 from fnmatch import fnmatch
@@ -186,7 +186,6 @@ class Command(BaseCommand):
                               "template %s\n" % template_name)
                 continue
 
-        contexts_count = 0
         nodes_count = 0
         offline_manifest = OrderedDict()
         errors = []
@@ -211,19 +210,14 @@ class Command(BaseCommand):
                         nodes_count += 1
                         template_nodes.setdefault(node, []).append(context)
 
-            threads = []
+            pool = concurrent.futures.ThreadPoolExecutor(max_workers=4)
             for template, nodes in compressor_nodes.items():
                 template._log = log
                 template._log_verbosity = verbosity
 
-                thread = threading.Thread(target=self._compress_template, args=(offline_manifest, nodes, parser, template, errors))
-                # Sticks the thread in a list so that it remains accessible and start thread
-                threads.append(thread)
-                thread.start()
+                pool.submit(self._compress_template, offline_manifest, nodes, parser, template, errors)
 
-            # Execute tasks in parallel
-            for thread in threads:
-                thread.join()
+            pool.shutdown(wait=True)
 
         # If errors exist, raise the first one in the list
         if errors:
@@ -236,7 +230,7 @@ class Command(BaseCommand):
 
         if verbosity >= 1:
             log.write("done\nCompressed %d block(s) from %d template(s) for %d context(s).\n" %
-                      (len(offline_manifest), nodes_count, contexts_count))
+                      (len(offline_manifest), nodes_count, len(contexts)))
         return offline_manifest, len(offline_manifest), offline_manifest.values()
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


karyon:
thank you very much!

it looks like this uses one thread per template, leading to potentially lots of threads running in parallel. this might result in high memory consumption as well as worse performance due to more overhead. also, while your case might be IO-bound, there are filters implemented in pure python, and then it would make sense to use processes again. in sum, i would suggest you use the `multiprocessing` module.

i would prefer to use locking. we have no guarantee that everything actually works if two threads compute the same thing, or whether there is some weird bug hidden somewhere, which might give us hard-to-debug problems down the line.

keithhackbarth:
@karyon -- Thanks for the feedback!

While I appreciate the recommendation, I'm not sure I understand the logic about why multiprocessing is better than threading for this case. Multiprocessing has much higher memory/resource overhead than threads do. 

I might challenge you that in all examples you mentioned: limiting memory consumption, pure python compressors, etc. that threads would outperform mulitprocessing on overall resource consumption and performance.

We run `manage.py compress` offline so multiple threads running in parallel aren't an issue, but if they were it would be fairly straightforward to just limit the number of threads.

keithhackbarth:
@karyon -- Thanks for feedback. I've updated pull request to using pool to limit the maximum amount of threads to 4. 

You can fairly easily switch between this and the multiprocess version by simply changing `ThreadPoolExecutor` to `ProcessPoolExecutor` and then making `offline_manifest` a pool variable.

I also tested both approaches on our codebase by switching from `uglifyjs` to  `rjsmin` and threading is still faster.

karyon:
it seems that this does not work on python 2.x. some googling turned up [this backport](https://pypi.org/project/futures/) of the futures module, which you might use, or you use the multiprocessing module. according to [this stackoverflow answer](https://stackoverflow.com/questions/3033952/threading-pool-similar-to-the-multiprocessing-pool), multiprocessing also has an (undocumented) implementation using threads.

karyon:
Python 2.7 support is dropped by now, so this doesn't need any larger changes for compatibility anymore. however, i would still like to have a fix for the race condition. while i don't mind duplicate work *that* much (though it's not nice either), i would definitely avoid modifying the dict with one thread while others might read from it, even if it most likely "just works" in current CPython versions.

to keep the lock time to a minimum, one option would be to write a dummy value into the dict right after the `if key in offline_manifest:` check, release the lock again, and then acquire it again to write the real computation result into the dict.

compressor/management/commands/compress.py
```diff
@@ -210,46 +210,55 @@ def compress(self, engine, extensions, verbosity, follow_links, log):
                         nodes_count += 1
                         template_nodes.setdefault(node, []).append(context)
 
+            pool = concurrent.futures.ThreadPoolExecutor(max_workers=4)
```
pawl-bb: It might be better to use `ProcessPoolExecutor` so you don't need to worry about thread safety?

Otherwise, you might need to add locks around the places where it manipulates shared objects (like `offline_manifest`).

It's probably a good idea to make the number of workers configurable too?

pawl-bb: Actually nevermind, there might still be some potential bugs when manipulating the shared objects with `ProcessPoolExecutor`.

You might need to refactor this to avoid that?

diox:
@keithhackbarth I know it's been a long time, and I'd completely understand if you don't care anymore, but if you want to take a look at @karyon's suggestion above and implement it, we could then merge this PR in time for the upcoming release.

keithhackbarth:
Thanks @diox. I've moved on to new projects so probably won't have time to work on this. Also, it has been a long time, but I still not sure if I understand the recommendation. The goal of my PR was the least wall-clock time. It seems like @karyon's goal is least CPU time. I value my personal time (wall-clock time) much more than CPU time which is why I never bothered implemented it.

karyon:
Hi! :)

I wasn't aiming for CPU time but for correctness. Currently there are threads simultaneously reading and writing to the same data structure. Since we don't know whether that is safe in Python, I was suggesting to implement something that's known to be safe.

diox:
I'd love for it to use locks to ensure peace of mind, but I don't have the time or energy to add this. @karyon do you ? If not, I think this can go in as-is.

karyon:
I don't, sorry :(

PR #926, opened 2019-01-03
Insert ';' after each js file instead of before to fix strict mode

Tasssadar:
The `;` at the start of compressed file breaks the JS strict mode. Append the `;` to the end instead to fix it.


karyon:
thanks. makes sense, and it's kinda stupid all the hashes need to be changed...

karyon:
if anyone volunteers, it would make sense to apply https://github.com/django-compressor/django-compressor/pull/931 at the same time.

karyon:
I cherry-picked this along with the hash changes from #951 in #952. thanks!

PR #928, opened 2019-01-28
Add log messages to debug missing keys.

xirdneh:
Debugging `OfflineGeneratorError` is quite difficult.
This PR adds some log messages that will show the strings used to create the offline manifest, hopefully this will give enough information so anybody can debug an `OfflineGeneratorError`.

Most of the time an `OfflineGeneratorError` shows up when compress gets a different string when creating the offline manifest and when a normal request happens. By printing the strings, key and template name that compressor uses and creates when building the offline manifest we give users enough information for them to realize what's the discrepancy between the manifest and a normal request and they can fix the error.

### DIFF
--- a/compressor/management/commands/compress.py
+++ b/compressor/management/commands/compress.py
@@ -235,7 +235,7 @@ class Command(BaseCommand):
                             settings.COMPRESS_URL, settings.COMPRESS_URL_PLACEHOLDER
                         )
                         offline_manifest[key] = result
-                        rendered_map[key] = {'rendered': rendered, 'template_name': template.name}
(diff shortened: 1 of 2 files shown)


karyon:
Thanks. I think we could make this more discoverable by adding a note to the output that was changed here: https://github.com/django-compressor/django-compressor/pull/853/files

Also, a test would be greatly appreciated. Since this is not covered at all, it might silently break in the future.

xirdneh:
@karyon Closing this since I don't have time to finish it.

karyon:
No problem. I'd leave it open though, maybe someone else finds time to finish this.

PR #931, opened 2019-03-29
Fix templates regarding w3c validator warnings. Removed type attribut…

akarca:
Fixed templates regarding w3c validator warnings. Removed type attribute from script element because the type attribute is unnecessary for JavaScript resources.


karyon:
looks good, thanks. unfortunately, a lot of tests would need to be updated. i won't have the time to do that anytime soon. if anyone volunteers, it would make sense to apply https://github.com/django-compressor/django-compressor/pull/926 at the same time.

karyon:
I cherry-picked this along with the hash changes from #951 in #952. thanks!

PR #934, opened 2019-04-12
Solves the issue 932

gh640:
I added the logic to check if the `settings.FILE_CHARSET` exists or not before accessing the value.

### DIFF
--- a/compressor/filters/base.py
+++ b/compressor/filters/base.py
@@ -119,7 +119,9 @@ class CompilerFilter(FilterBase):
     """
     command = None
     options = ()
-    default_encoding = settings.FILE_CHARSET
+    default_encoding = (
+        settings.FILE_CHARSET if hasattr(settings, 'FILE_CHARSET') else 'utf-8'
+    )
 
     def __init__(self, content, command=None, **kwargs):
         super(CompilerFilter, self).__init__(content, **kwargs)
--- a/compressor/filters/base.py
+++ b/compressor/filters/base.py
@@ -120,7 +120,8 @@ class CompilerFilter(FilterBase):
     command = None
     options = ()
     default_encoding = (
-        settings.FILE_CHARSET if hasattr(settings, 'FILE_CHARSET') else 'utf-8'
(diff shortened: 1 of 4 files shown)


compressor/filters/base.py
```diff
@@ -119,7 +119,9 @@ class CompilerFilter(FilterBase):
     """
     command = None
     options = ()
-    default_encoding = settings.FILE_CHARSET
+    default_encoding = (
+        settings.FILE_CHARSET if hasattr(settings, 'FILE_CHARSET') else 'utf-8'
```
karyon: does that work? i just fired up a console and the `hasattr` returns true on django 2.2. You could experiment with  `settings.is_overridden`, perhaps `settings.FILE_CHARSET if settings.is_overridden('FILE_CHARSET') else 'utf-8'` might work?

compressor/tests/test_filters.py
```diff
@@ -31,16 +31,20 @@ def blankdict(*args, **kwargs):
 
 @override_settings(COMPRESS_CACHEABLE_PRECOMPILERS=('text/css',))
 class PrecompilerTestCase(TestCase):
+    CHARSET = (
+        settings.FILE_CHARSET if hasattr(settings, 'FILE_CHARSET') else 'utf-8'
+    )
```
karyon: i think for the tests (here and in that other file), we can indeed just replace every FILE_CHARSET with utf-8. i mean we're reading our own content here, and that should be utf-8...

compressor/filters/base.py
gh640: Thanks for the review. You're right. The check with `hasattr()` didn't work properly and I replaced it with `is_overridden()` with the new commit. It seems to work well now.

compressor/tests/test_filters.py
gh640: I also replaced them to just `'utf-8'` in the new commit.

gh640:
@karyon thanks very much. I fixed the points and I'd like you to review this again.

karyon:
cool, thanks!

gh640:
Thanks!

PR #942, opened 2019-06-04
Fix release date of v2.3

jodal:


diox:
Thanks, knew I would screw up *something* :)

Issue #946, opened 2019-06-26
Usage when also using django-csp


darakian:
Hey all,

I've got a bunch of js files in html tempaltes which I'm compressing as
```
{% compress js %}
    <script src="{% static "dir1/code1.js" %}"></script>
    <script src="{% static "dir2/foo/bar/js/code2.js" %}"></script>
    <script src="{% static "dir2/foo/bar/js/code3.js" %}"></script>
    <script src="{% static "dir2/foo/bar/js/code4.js" %}"></script>
...
{% endcompress %}
```
And this works well. However I'm trying to add nonce-ing as 
```
{% compress js %}
    <script nonce="{{request.csp_nonce}}" src="{% static "dir1/code1.js" %}"></script>
    <script nonce="{{request.csp_nonce}}" src="{% static "dir2/foo/bar/js/code2.js" %}"></script>
    <script nonce="{{request.csp_nonce}}" src="{% static "dir2/foo/bar/js/code3.js" %}"></script>
    <script nonce="{{request.csp_nonce}}" src="{% static "dir2/foo/bar/js/code4.js" %}"></script>
...
{% endcompress %}
```
The string `{{request.csp_nonce}}` seems to get picked up as a jinja variable and I'm getting a `OfflineGenerationError: You have offline compression enabled but key "some hash" is missing from offline manifest.` I've tried adding 
```
COMPRESS_OFFLINE_CONTEXT = {
        'request.csp_nonce': '{{request.csp_nonce}}',
    }
```
to my settings file, but no dice. Has anyone worked with this combination? Am I missing something obvious?

diox:
There are a couple issues here:
- `{{ request.csp_nonce }}` is dynamic, so you can't easily have it in `COMPRESS_OFFLINE_CONTEXT`.
- Even if you did, compressor's uses its own templates for generating the `<script>` so it would just not be used at all.

The way out of this is to modify the template compressor uses (`compressor/js_file.html` or `compressor/js_inline.html`, include your own in one of your template dirs making sure it's overriding compressor's) to make use of `{{ request.csp_nonce }}`. Essentially doing it that way completely bypasses the offline context problem

Ideally, someone (you? :) would make a patch against compressor that does this automatically if a `request_csp_nonce=True` parameter is passed and django-csp is installed, it could be useful to others.

darakian:
[...]

PR #947, opened 2019-06-26
Add CSP nonce support to compressed JS and CSS in online context

darakian:
This is a very work in progress PR as I'm not sure where everything in the codebase is, but this PR comes from https://github.com/django-compressor/django-compressor/issues/946

The idea is to have two user configurable variables which will be placed as is inline with the script/style tags. ie `INLINE_SCRIPT_TAG='foobar123'` would result in scripts starting with `<script foobar123 src=...</script>`. The primary motivator is to allow compressor to work easier with the `django-csp` nonce mode.
https://django-csp.readthedocs.io/en/latest/nonce.html

I'm sorry, but at the moment I'm not sure how to pass the variables defined in conf into the filter function. If someone can point me at an example I'd be happy to update this PR.


diox:
I'd prefer to see direct implementation of django-csp without being super generic tbh. I don't like the idea of directly passing attributes, and doing it from a configuration variable seems convoluted - you have to pass an unrendered string, which feels all kinds of wrong.

darakian:
Ok, so would you prefer two boolean variables which pass the explicit variable? I've updated the PR to reflect this, but again the linkage to the templates is unclear.

diox:
Shouldn't need a configuration variable at all. It's less trivial than I thought however, I had forgotten how some of the internals of offline compression works. I'm pasting my whole thought process, dead-end included, so you can see the reasoning.

Ideally, you'd just pass `nonce={{ request.csp_nonce }}` in your own template in the HTML surrounded by `{% compress %}`, and compressor would just pass that down to its own templates. Sadly, that wouldn't work with offline compression, because compressor wouldn't be able to generate the right value for this variable when you run the `compress` command. So, instead:

- Modify both the django and jinja template tags to accept a `django_csp_nonce` boolean parameter. It should be stored on the `CompressorNode` and `CompressorExtension` instances, respectively. That'll make the template tag know whether or not to include the nonce from django-csp.
- Speaking of django-csp, add a check to raise an exception if `django_csp_nonce` was passed to the template tag and django-csp isn't installed
- In  `compressor.templatetags.compress.CompressorMixin.render_compressed()`, set `django_csp_nonce` in context from the value of the property stored on the current instance
- In the templates, add something like:
`{% if django_csp_nonce %}nonce="{{ request.csp_nonce }}"{% endif %}`. (`request` should already be in the context, I think)

Unfortunately that still wouldn't work with offline compression, because the whole thing would only be rendered once, during the `compress` management command step. The idea still makes sense, but instead of adding the dynamic bit to the templates (which are only rendered once) we need to generate the template in a way that makes it easy to add the nonce afterwards. Something like:
- In the templates, instead of doing the bit earlier, add something like `{% if django_csp_nonce %}nonce="__django_compressor_csp_nonce"{% endif %}`
- In `compressor.templatetags.compress.CompressorMixin.render_compressed()`, refactor the method to only have a single `return` statement, and before returning the value, regardless of how it was generated, add a `.replace('__django_compressor_csp_nonce', context['request'].csp_nonce)` if `django_csp_nonce` is true in context.


Make sure it works, add some tests and docs covering how this works and hopefully you're done :)

darakian:
Updated to be along those lines. I suppose it's worth considering styles and scripts together as a single case. Where does context originate from?

darakian:
@diox Perhaps this would be better simply as a feature request.

darakian:
@diox Ok, I think I understand what's going on in this library now. I've got some code which passes testing, but as I understand it this method pulls in the csp_nonce value directly rather than filling in the value which django_csp would fill in.
ie.
`{% if django_csp_style_nonce %}nonce="__django_compressor_csp_nonce" {% endif %}` renders to `nonce=12345` rather than `nonce={{request.nonce}}`.
If that is the case then I'm stalled again on how to write a test for this. The tests I see are overriding settings values rather than context values. Perhaps I'm missing something, but if you could chime in that would be helpful.

darakian:
@diox ping

compressor/templatetags/compress.py
```diff
@@ -90,36 +90,48 @@ def render_cached(self, compressor, kind, mode):
...
+
+        if compressed_content is None:
+            file_basename = name or getattr(self, 'basename', None)
+            if file_basename is None:
+                file_basename = 'output'
+            compressed_content = compressor.output(mode, forced=forced, basename=file_basename)
+        # Do the CSP nonce replacement
+        if context.get('django_csp_nonce'):
+            compressed_content = compressed_content.replace('script ', 'script nonce={{request.nonce}} ')
+            compressed_content = compressed_content.replace('type="text/css"', 'type="text/css" nonce={{request.nonce}} ')
```
diox: Sorry, these 2 `replace` calls are super dodgy. Why didn't you follow my suggestion to add a specific string we could look for in the template and search&replace *that* instead?

More importantly, I don't see how that helps. You now are outputting HTML that contains `nonce={{request.nonce}}`, which is then not parsed by the template engine, so the litteral nonce you get is `{{request.nonce}}` and not the actual nonce value.

This is why I suggested to replace directly with the `context['request'].csp_nonce` value.

darakian: I agree they are dodgy, I was just trying to get something to pass testing and I thought this might be equivalent. I suppose not. I've push an update to the PR which reintroduces a tag and looks for that as a replacement. I was unaware that the output would not be revisited by the django csp plugin. 

I'm still unsure of how to force the if case in the template to be true during testing.

darakian:
All good. I figured you might be away :)

darakian:
@diox ping

diox:
Can you fix the conflicts so that travis can run on the PR ?

darakian:
Updated

darakian:
@diox ping

diox:
I haven't tested it but this looks reasonable to me now. Is it working for you like you wanted?

darakian:
@diox I'm hitting an error with offline compression of the sort   
`ValueError: Missing staticfiles manifest entry for 'foo/bar/baz/some.js'`  
I think the issue is that the key `__django_compressor_csp_nonce` is undefined in the offline context. I'm going to try replacing `__django_compressor_csp_nonce` with the string `nonce={{ request.csp_nonce }}` so that when the static file is served the django_csp lib will render the nonce into it. If you have any thoughts on this approach please let me know.

DylanYoung:
I'm curious.  Wouldn't it make sense to ignore *all* undefined template variables (those not in OFFLINE_CONTEXT) when running `./manage.py compress` or at least have a specific whitelist?  That way compressed blocks can still contain other templated variables tied to the request cycle.  I think then the only work is lifting these variables to the new tags created by compressor.

DylanYoung:
To clarify/elaborate:

Instead of templating a new js/css block out, you simply identify the start and close tags, render them with OFFLINE_CONTEXT (preserving any template blocks containing undefined variables... this might be the sticking point...), and then allow them to be rendered as usual in the request-response cycle. 

Some sticking points:
1) How do you "merge" multiple blocks that are being combined? Probably we need to verify that they are "compatible" probably by an exact match between attrs (though this could be configured). 
2) *Can* we do a partial render of templates with both supported engines (Jinja and Django)?

darakian:
@DylanYoung I think ignoring all undefined tags makes perfect sense. Though I'm not familiar enough with where the compression process to say how to go about doing it. I'm not sure if @diox has any input on that idea, but perhaps a separate PR should be created for that feature.

diox:
Getting both supported template engines (or even one) to ignore variables during the compression process would be tricky. Maybe not impossible - it's been a while since I looked at that code, I don't use compressor anymore day to day - but certainly not trivial.

darakian:
Hmmm fair enough. @diox In that case would you mind if I merge this PR an open a separate one to work on ignoring variables?

DylanYoung:
Merge this one for now if you think it's up to snuff,  but I have an idea:
What if compressor includes template tags/filters for inserting the style
or script nodes themselves. It could be added as a method in addition to
current functionality. This would also add a hook for people to create
their own variations, one that adds the django-csp nonce automatically,
for example.   Hmmmm... I need to think more on it, but I think it might
work.

The implicit functionality would be slicker for sure,  but also more
difficult and error prone without proper api support from jinja and django.

DylanYoung:
Or even arguments to the compress tag itself: script_tag_template,
style_tag_template?

Very rudimentary,  but super simple to implement and extremely flexible.

diox:
If you have a concrete proposal to make, please file an issue describing it in detail, it's starting to be a little off-topic here. But do note that the implementation will then likely have to come from you as well :)

darakian:
@diox Mind if I just make an issue to discuss the changes? Also, do you have any advice on the codecoverage thing blocking this PR?

diox:
You can ignore coverage issues for now.

I'm sorry but I'm afraid I don't have much more guidance to offer - compressor is a relatively stable project, I haven't used it in years, so I can review pull requests and give some insight but you'll have to do the heavy lifting if you want any additional functionality.

The PR seems OK to me, and I'm willing to merge it /if it solves your problem/. If it's not enough to solve it, then I don't see the point of merging right now.

darakian:
> You can ignore coverage issues for now.

Ok cool. Fair enough on the lack of guidance and this PR does not solve my issue (at least not completely). I think the correct approach is for me to work on the offline context and solve that before revisiting this PR. Thanks for taking the time to work through this with me :)

As a next step I'll make an issue to discuss with jinja tag parsing and go from there.

string-areeb:
@diox When generating and checking the hash for offline compression, we render the content and then hash it. What is the problem in hashing before rendering, so instead of hashing 

```html
<script nonce="435kjfs32">media/lol.js</script>
```

we hash:

```html
<script nonce="{{ request.csp_nonce }}">{% static 'lol.js' %}</script>
```

This will solve our problem of offline compression. Is there any issue in this?

diox:
If the input was this:
`<script src="{{ foo }}"></script>` you don't want to hash `<script src="{{ foo }}"></script>`, you want to resolve `foo` so that the proper compressed content is generated by offline compression.

string-areeb:
But that would anyway be generated by the compressor, the only change would be the key of the generated content, which is the hash

diox:
I don't understand what you're proposing. We need to generate compressed content for each possible value of `foo` when running the command, and then figure out the right one to return during the request/response cycle. If you hash the content before rendering, you don't distinguish between every possible value of `foo` and therefore you can't return the right one ?

string-areeb:
I was assuming that foo is idempotent. If it isn't then yes, ut won't work

iamareebjamal:
@diox I have added a very crude POC in #989 which works with offline compression. Several things remain to be added, but can you please see if the approach is fine by you

PR #950, opened 2019-07-06
Use six directly instead of django.utils.six

### LINKED ISSUE #949: Django will drop django.utils.six in 3.0
The vendored `django.utils.six` is scheduled to be removed (well, already removed in `master`) in preparation of the upcoming Django 3.0 release.

Unless you want to drop Python 2 and all Django versions still supporting Python 2, it's probably easiest to just depend on `six`.

rixx:
Closes #949

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -144,5 +144,6 @@ setup(
         'django-appconf >= 1.0',
         'rcssmin == 1.0.6',
         'rjsmin == 1.1.0',
+        'six == 1.12.0',
     ],
 )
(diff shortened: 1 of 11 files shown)


karyon:
thanks!

setup.py
```diff
@@ -144,5 +144,6 @@ def find_package_data(where='.', package='',
         'django-appconf >= 1.0',
         'rcssmin == 1.0.6',
         'rjsmin == 1.1.0',
+        'six == 1.12.0',
```
jaap3: Why is this pinned to 1.12.0? 1.13.0 has been release recently.

rixx: Because at the time of this PR, 1.13.0 had not been released yet. The release date was November 5th, some months later.

Issue #690, opened 2015-11-02
Support additional attributes in script tag, e.g., crossorigin="anonymous"


mrcoles:
It would be great to have support for additional attributes in the generated script tag from something such as:

```
{% compress js %}
<script src="foo.js"></script>
<script src="bar.js"></script>
{% endcompress %}
```

to support things like `crossorigin="anonymous"`—which for this particular example gives certain privileges to listen to things like `window.onerror`.

I wonder if it could be something like `{% compress js crossorigin="aynonymous" %}` or settings parameter that allows for setting additional script tag attributes?

In the meantime, I’ve overwritten the template in my local templates directory in my project:

```
$ cat templates/compressor/js_file.html
<script crossorigin="anonymous" src="{{ compressed.url }}"{{ compressed.extra }}></script>
```

martyzz1:
I was also just wondering how to achieve this..  thanks for the template tip

anuragjain:
I am also stuck with the same problem. Can you please explain how did you get the {{ compressed.url }}?
I couldn't find anything in the docs as well. 
Thanks in advance

martyzz1:
within the compressor code  there is a templates directory, which compressor uses to write the script html. e.g. lib/python2.7/site-packages/compressor/templates/compressor 

I simply created a local file 

project/templates/compressor/js_file.html

who's contents are
    <script defer type="text/javascript" src="{{ compressed.url }}"{{ compressed.extra }} crossorigin></script>

And this worked for me

anuragjain:
Awesome, thanks for your help.
I was always looking in the compress template tag and got confused there, your solution worked perfectly.

martyzz1:
no worries... just pieced together advice from others:)

jessemorton:
Thanks @martyzz1 @mrcoles

andypillip:
This would be also useful/necessary for CSS.

It's state of the art to defer CSS loading, which is possible by means of the `rel="preload" as="style".

But using `{% compress css %}` will remove the attributes.

nordigen-account:
This also affects us. We want to add `crossorigin` so that we can get the full exception messaging when serving from our CDN. Right now we can't see production errors from our own scripts.

PR #951, opened 2019-07-22
Add support for rel=preload

jsumnerPhD:
Hi all,

This is a small change to simplify generating the `<link rel="preload">` tag with the correct hash for a given css or js file.

Tests and docs included.

Cheers!

### DIFF
--- a/compressor/base.py
+++ b/compressor/base.py
@@ -349,6 +349,13 @@ class Compressor(object):
         """
         return self.render_output(mode, {"content": content})
 
+    def output_preload(self, mode, content, forced=False, basename=None):
+        """
+        The output method that returns <link> with rel="preload" and
+        proper href attribute for given file.
+        """
+        return self.output_file(mode, content, forced, basename)
+
     def render_output(self, mode, context=None):
         """
         Renders the compressor output with the appropriate template for
--- a/compressor/contrib/jinja2ext.py
+++ b/compressor/contrib/jinja2ext.py
@@ -49,7 +49,7 @@ class CompressorExtension(compress.CompressorMixin, Extension):
                 # The file mode optionally accepts a name
                 if parser.stream.current.type != 'block_end':
                     namearg = const(parser.parse_expression())
-            elif modearg.value == compress.OUTPUT_INLINE:
+            elif modearg.value == compress.OUTPUT_INLINE or modearg.value == compress.OUTPUT_PRELOAD:
                 pass
             else:
                 raise TemplateSyntaxError(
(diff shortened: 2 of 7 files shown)


jsumnerPhD:
Updated the test suite for #931 and #926.  

Hard-coded hashes updated by working through failing unit tests one by one... maybe there was an easier way?

Cheers!

karyon:
I cherry-picked the second commit into #952 and merged that. thanks!

regarding the first commit, i don't fully understand the matter, but it seems there are other attributes besides preload that could be supported, see #690. if that's the case, it seems like your solution does not scale to multiple attributes because you'd need one OUTPUT_MODE per attribute combination, is that correct?

jsumnerPhD:
Hi karyon,

`rel=preload` is more like a directive than an attribute: it is intended to give the browser a heads up that a resource will be needed, so start loading it now.  The purpose of adding a `preload` output mode to django-compressor is to be able to take advantage of this directive for our own cached resources.

With this additional output mode, django-compressor can now generate an `inline` compressed script/stylesheet, an external compressed `file`, or simply the `preload` tag (which does the same thing as `file` but returns the preload tag instead of the script/link tag).  It really is unrelated to the problem of extra attributes.

My workflow with this version of django-compressor uses `preload` mode in the head and then calls the actual scripts as needed.  `preload` really just takes care of generating the correct filename in the tag: instead of `<script src="...">` in the head, it will output `<link rel="preload" as="script" href="...">`.

```python
<head>
...
{% compress js preload %}
<script src="{% static 'website/js/index.min.js' %}"></script>
{% endcompress %}
...
</head>

<body>
...
{% compress js %}
<script src="{% static 'website/js/index.min.js' %}"></script>
{% endcompress %}
...
</body>
```
Cheers!

karyon:
thanks!

timonweb:
Any idea when this is going to be released?

karyon:
no, sorry.

(fwiw, it's really really easy to put a specific commit of a git repository into your requirements.txt, try it! :D there's no additional testing or the like done on releases, so there's little downside for you.)

diox:
I'm planning to do a release later this month. Please use the develop branch in the meantime.

VivienGiraud:
Nice, but it seems that it's not working when COMPRESS_ENABLED is False, then it juste leaves the js files ready to be executed :/

karyon:
@VivienGiraud can you please open a new issue with the actual and expected behavior?

VivienGiraud:
@karyon Thanks, here is the link : https://github.com/django-compressor/django-compressor/issues/987

PR #954, opened 2019-08-13
support cloud completely

tuan-nng:
use default-storage to store file instead of using FileSystemStorage

We will not support python 2.7

### DIFF
--- a/compressor/base.py
+++ b/compressor/base.py
@@ -15,7 +15,7 @@ from compressor.conf import settings
 from compressor.exceptions import (CompressorError, UncompressableFileError,
         FilterDoesNotExist)
 from compressor.filters import CachedCompilerFilter
-from compressor.storage import compressor_file_storage
+from compressor.storage import compressor_file_storage, default_storage
 from compressor.signals import post_compress
 from compressor.utils import get_class, get_mod_func, staticfiles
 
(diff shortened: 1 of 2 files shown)


karyon:
Can you give any more detail what is broken/suboptimal right now and what this patch improves exactly?

Also, since django 1.11 (which supports python 2.7) is supported until april 2020, i'd rather not break python 2.7 support now already.

tuan-nng:
when the cache checks the modified time of a file, it calls os and checks against local file system. This makes this library impossible to use cloud storage as a backend storage completely. 

This PR tries to get that check by calling default_storage module.

tuan-nng:
> Also, since django 1.11 (which supports python 2.7) is supported until april 2020, i'd rather not break python 2.7 support now already.

Yes, I will not merge this PR. Close this for now

karyon:
I'd keep this open, I can wait half a year :)

karyon:
we've dropped support for python 2.7, so this could be merged.

however, i have tried to understand the code surrounding this and it seems like this only applies if offline compression is not enabled. So if you used a cloud storage for your static assets AND don't use offline compression, doesn't that imply compressor potentially downloads assets from your cloud storage, processes them, and uploads them again all while some client is requesting a page? why would you want to do that? While I'm not going to stop you, I'm trying to understand the use case this change is enabling :)

for the protocol, i checked whether the other .opens in the codebase need this change as well. the only other relevant one was in CompilerFilter.input, and that one handles the output of calls to local executables, which i guess always output local files.

PR #957, opened 2019-10-12
Add support for the Calmjs javascript minifier

alvra:
Calmjs is a javascript minifier similar to SlimIt that has some advantages. https://github.com/calmjs/calmjs.parse
For me, it solves the following problem:
https://github.com/rspivak/slimit/issues/97


diox:
Hi,

Sorry, but I think this doesn't merit inclusion in django-compressor core. You can use it in your own project if needed as it doesn't require anything special on compressor's side to do that.

alvra:
SlimIt hasn't seen a new release since 2013, which means right now, there are a number of problems with it. For example, pretty much every feature of ES6 is unsupported. Pull requests for these features have been waiting since as far back as 2013.

Meanwhile Calmjs' last release was a few months ago, ES6 is partially supported, it can generate source maps, and it's slightly faster than SlimIt.

If Django-Compressor is to provide a single js minifier out of the box, I don't think SlimIt is the right choice anymore IMO. We should keep it for backwards compatibility, but please consider adding a better option like Calmjs.

diox:
Thanks for your reply, I appreciate it. I don't really like shipping too many filters by default, especially relatively trivial ones, but I you make a good point about slimit not cutting it any more.

diox:
One thing I don't like about the slimit integration, and this one has the same problem, is that there isn't even a basic test covering it. Could you add the latest `calmjs` version in `requirements/tests.txt` and add a test making sure it works? You can draw inspiration from the `JsMinTestCase` one in `compressor/tests/test_filters.py`.

alvra:
Thanks! It bothered me too, so I added tests for both SlimIT and Calmjs.

diox:
Thank you for your contribution! I'll do a release eventually later this month, in the meantime it'll be available on the develop branch.

alvra:
Great, thanks!

PR #964, opened 2019-12-03
Release v2.4

albertyw:
For #963  .  Depends on #966.

 I took a stab at the changelog too.  Feel free to modify before merging.

I was a bit confused with #957 because it pulls in several commits:
```
| * ca7b4c3 - Merge (7 weeks ago) <Alexander van Ratingen>
| * 9040dbc - Update mention in documentation of old filter setting to new form. (1 year, 8 months ago) <Alexander van Ratingen>
| * 0ff2cff - Document new COMPRESS_FILTERS setting. (1 year, 8 months ago) <Alexander van Ratingen>
| * f3cb088 - Remove old COMPRESS_FILTER_CSS/JS settings from configuration after converting them into the new COMPRESS_FILTERS[css/js] form. (1 year, 8 months ago) <Alexander van Ratingen>
| * 0a3b5b5 - Process review comments (1 year, 9 months ago) <Alexander van Ratingen>
| * 2a8ab45 - Fix failing test on Python 2. (1 year, 9 months ago) <Alexander van Ratingen>
| * eca31d1 - Abstract away resource types to allow the implementation of new types. (1 year, 9 months ago) <Alexander van Ratingen>
```
But the changes in these commits seem to have been reverted before merging back on the `develop` branch.  

After this is merged, someone with permissions will still need to `python setup.py sdist bdist_wheel` and upload to pypi.

PR #965, opened 2019-12-03
Remove support for python 2.7, 3.4 and django 2.1

albertyw:
Depends on #966 

Drop support for python 2 - EOL January 1, 2020
Drop support for django 2.1 - EOL December 2019
Drop support for python 3.4 - EOL March 2019


karyon:
django 1.11 is supported until april 2020 and supports python 2.7 and 3.4. Personally i would either support django 1.11 and all the respective python versions, or not support django 1.11 at all. 

I would be fine with dropping support for django 1.11 and python 2.7 and 3.4 ahead of schedule, but then i'd also vote for calling it 3.0 so we can simply say "django-compressor 2.x is the last version to support all of that".

diox:
Yeah I also commented that I'm in favor of naming it 3.0 in #966. I don't mind Dropping Django 1.11 for that.

albertyw:
Python 2 is EOL earlier than django 1.11 so I think python 2 support should be dropped now and django 1.11 later.  

I can also split this PR to have one PR that adds support for Django 3 and Python 3.8 and another PR that drops support for python 2, django 2.1, and python 3.4.  See also https://github.com/django-compressor/django-compressor/pull/964#issuecomment-561275170

karyon:
I understand, but I disagree. Not supporting the python versions that django 1.11 supports means that we're not fully supporting django 1.11. There's no need for us to make it harder than necessary for downstream developers to figure out which django and python versions they can use, and we can make it simpler by sticking to the policy of our "parent" project as we've done in the past. Also, dropping python 2.7 support now and django 1.11 support later would mean more major releases with breaking changes for us.

So, I'd prefer to do a release now without breaking changes, and doing a 3.0 release without django 1.11 support, without python 2.7 support, and without six later (when it's done at the earliest :) ). @diox, I guess in the end it's your call since you're actually doing it :)

diox:
Yep, let's not make it more complex that it needs to be. I don't mind fully dropping 1.11 and Python 2 together in our 3.0 release, it'll make things more straightforward, and we're going to do a 2.4 release that keeps that compatibility before anyway.

albertyw:
I'm under the impression that django 1.11 implicitly drops support of python 2 when python 2 stops being supported itself, because python is the "parent project" of django.  I definitely get not having a lot of version churn just to add/drop support for dependencies of django-compressor though.  

I'll leave the decision to @diox though.  I can split this PR.

jrief:
@karyon With Django-3.0 I removed Python-2.7 support from my many other projects but not for Django-1.11. Django-1.11 is a LTS and still supported until 2020/Q2, so that IMO should be kept.

In practice, we removed all occurrences of `six`, removed the `from __future__ import unicode_literals` and fix class inheritance and the super-calls. Django-1.11 had no problems with that, so in my opinion it would be safe to support Django-1.11–3.0 and Python-3.5–3.8 with the exception of Python-3.5+Django-3.0.

debdolph:
Any progress on this?

karyon:
see https://github.com/django-compressor/django-compressor/pull/964 for the plan (that PR is also the next one to be merged. do you need this PR for anything in particular?

debdolph:
django-compressor is a dependency of django-dbbackup :).

PR #967, opened 2019-12-04
Test against python 3.8 and django 3.0

albertyw:
This is a subset of #965 that keeps backwards compatibility by only adding support for django 3.0 and python 3.8, instead of removing support for older versions of django and python.  

Add support for python 3.8 - Released October 2019
Add support for django 3.0 - Released December 2019

Prerequisite for #964

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,8 +40,6 @@ matrix:
     - python: 3.8
       env: TOXENV=py38-2.2.X
       dist: xenial
-    - python: 3.5
-      env: TOXENV=py35-3.0.X
     - python: 3.6
       env: TOXENV=py36-3.0.X
     - python: 3.7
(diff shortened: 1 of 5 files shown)


albertyw:
So, django-sekizai depends on django-classy-tags, which doesn't have a release which is django 3.0 compatible.  There's a PR on the latter to support django 3.0 (https://github.com/divio/django-classy-tags/pull/46), but hasn't had movement in over 2 months.  

Not sure how we want to proceed.  We could have django-compressor pin a fork of django-classy-tags but that seems unsafe.  We could also drop django-sekizai, but that would be backwards incompatible

karyon:
Thanks! with regards to django-classy-tags, i left them a review and a comment. Lets see whether there's a response within, say, a week or so. If they don't move, I'd tend towards using a fork, while it's not elegant I think it's still fine since we're using it strictly for testing only.

albertyw:
I fixed a deprecation.  This should be ready to merge now.  

(codecov is saying that there's an issue with one of the python files decreasing coverage but that's only because it's not reading the coverage from the django 3.0 test)

PR #968, opened 2019-12-04
Update flake8 to be compatible with newer versions of python

albertyw:
This is a requirement for #967.  Python 3.8 breaks flake8 3.5.0, so upgrading flake8 to 3.7.9 makes testing compatible with python 3.8, but it drags along a bunch of new flake issues, particularly the fun style flip-flop from https://lintlyci.github.io/Flake8Rules/rules/W503.html.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ testenv:
 	pip install Django
 
 flake8:
-	flake8 compressor --ignore=E501,E128,E701,E261,E301,E126,E127,E131,E402
+	flake8 compressor --ignore=E501,E128,E701,E261,E301,E126,E127,E131,E402,W503
 
 runtests:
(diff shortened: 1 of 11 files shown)


karyon:
help me with this, it seems like W504 is active, otherwise you wouldn't have made those fixes i guess. but you're ignoring W503 here - does that mean they are both active by default??

didn't know about the discussions around that part in PEP8, that was an interesting read, thanks!

albertyw:
I see that codecov is failing, but can we get this merged?  I don't think linting changes with no logical changes warrants backfilling a bunch of tests.

diox:
Yeah that's fine. Eventually we may want to adopt Black and get rid of all these ignores...

PR #972, opened 2019-12-09
Allow six to be upgraded

jaap3:
I see there's an ongoing discussion about dropping Python 2 support altogether, so feel free to ignore this PR if that's the way forward.


karyon:
Thanks!

For the protocol, I briefly contemplated whether it's a good idea to have an open version range in there, but given the scope and stability of the six project, that shouldn't cause issues, also we'll remove this for 3.0 anyway. (still, it makes sense for our 2.x release!)

PR #977, opened 2019-12-16
Switch from smart_text to smart_str

albertyw:
`smart_text` is deprecated in favor of its alias, `smart_str` in Django 3.0 - https://docs.djangoproject.com/en/3.0/ref/utils/#django.utils.encoding.smart_text

Even in django 2.0, the two functions were aliases of each other (just in the opposite direction) when used for python 3.  

Depends on #965


karyon:
thanks!

karyon:
there seems to be a behavior change in django 1.11 and python 2.7, the tests are failing there. so this will have to wait until we remove support for those.

albertyw:
This PR is meant for after #965 is landed, for django-compressor v3.0.  Sorry I didn't make that clear.

Issue #973, opened 2019-12-11
compress tags not found in {% elif %} tags


nwilson5:
I was having issues with offline manifest missing hash keys. Turned out, if {% compress %} was nested in an {% elif %} tag it was not being caught.

I believe this can be rectified by updating compressor.offline.jinja2.Jinja2Parser's get_nodelist function.

https://github.com/django-compressor/django-compressor/blob/c99f0c27003303658ea3e652c8f068bf698363b9/compressor/offline/jinja2.py#L107-L113

Fix for my purposes:
```
    def get_nodelist(self, node):
        body = getattr(node, "body", getattr(node, "nodes", []))

        if isinstance(node, jinja2.nodes.If):
            return body + node.elif_ + node.else_

        return body

```
Specifically: return body **+ node.elif_** + node.else_

Making that change on my source code seemed to fix my issue with it. Not familiar enough with django-compressor to know if there are any issues making this change.

karyon:
Seems very reasonable. thanks for the research! could you open a PR with that exact change?

You'd get bonus points for a test demonstrating the problem :D
I think that should be relatively easy. we already have a test that tests a compress tag inside an `if` tag (OfflineCompressConditionTestCase). you could extend the template (located in tests/test_templates[_jinja2]/test_condition) by an elif clause, with some other nonsense javascript in there, just like the if clause already contains. then you'd copypaste `OfflineCompressConditionTestCase` and put something false-y into the condition variable.

nwilson5:
@karyon, not entirely certain what you are asking re: the test but I made a pull request and attempted to make a test for the elif tag. Not written a test before, so let me know if that's not what you meant.

nwilson5:
@karyon it appears the test I added is failing in some cases. I was running it in my environment OK.
The Jinja2 requirement is the appropriate one for including node _elif tag. I feel my code change is appropriate, but perhaps my test is not.

PR #981, opened 2019-12-24
jinja2 offline compression bug

nwilson5:
Re: https://github.com/django-compressor/django-compressor/issues/973
Jinja2 added _elif, compress inside an elif tag getting missed in offline compression.

### DIFF
--- a/compressor/__init__.py
+++ b/compressor/__init__.py
@@ -1,2 +1,2 @@
 # following PEP 386
-__version__ = "2.2"
+__version__ = "2.3"
--- a/compressor/offline/jinja2.py
+++ b/compressor/offline/jinja2.py
@@ -108,7 +108,7 @@ class Jinja2Parser(object):
         body = getattr(node, "body", getattr(node, "nodes", []))
 
         if isinstance(node, jinja2.nodes.If):
-            return body + node.else_
+            return body + node.elif_ + node.else_
 
         return body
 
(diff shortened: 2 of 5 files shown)


nwilson5:
Not super familiar with writing tests, feel free to change it as you see fit.

karyon:
Thanks! how did you run the tests? specifically, it's rather important that you have requirements/tests.txt installed and use compressor/test_settings.py as settings.

also, your PR contains some old commits. That doesn't matter much though, we can just squash-merge the PR so those will be gone.

karyon:
also, please verify that the output files (which are pointed at by the new script tags) contain what you expect.

nwilson5:
> Thanks! how did you run the tests? specifically, it's rather important that you have requirements/tests.txt installed and use compressor/test_settings.py as settings. also, please verify that the output files (which are pointed at by the new script tags) contain what you expect.
> 
> also, your PR contains some old commits. That doesn't matter much though, we can just squash-merge the PR so those will be gone.

I'm unfamiliar with writing tests or contributing to django-compressor specifically, so not sure what actually needs to be done here. I am not interested in figuring out how to create a viable pull request to fix this issue. I've forked and fixed this for my needs. It is a bug that needs to be fixed, and it's a fairly simple fix, so if the contributors to this package feel it worthwhile you're welcome to sort out the tests.

karyon:
for the protocol, the change seems correct, but the newly added tests are failing. if anyone wanto to pick this up, you'd need to debug the test and verify that the rendered output of `_render_script` and `_render_result` in `_test_offline` look good, and then adjust the hashes.

PR #982, opened 2019-12-31
Remove python 2 logic

albertyw:
Depends on #976


karyon:
Thanks @albertyw for all your work! I have created a new [milestone for 3.0](https://github.com/django-compressor/django-compressor/milestones/3.0), your input both in code and what else should be done is appreciated!

PR #988, opened 2020-01-23
Replace force_text with force_str

joren485:
### Description

This PR fixes the following deprecation warning:
```
RemovedInDjango40Warning: force_text() is deprecated in favor of force_str()
```
`force_text` is deprecated in Django 3 and should be replaced with `force_str` ([release notes of Django 3](https://docs.djangoproject.com/en/dev/releases/3.0/#features-deprecated-in-3-0)).


karyon:
thanks!

PR #989, opened 2020-02-17
Add Django CSP nonce support

### LINKED ISSUE #946: Usage when also using django-csp

iamareebjamal:
- [x] Add template changes
- [x] Add branches for non-csp users
- [x] Check offline compression for every case

Fixes #946

### DIFF
--- a/compressor/conf.py
+++ b/compressor/conf.py
@@ -82,6 +82,8 @@ class CompressorConf(AppConf):
     # Placeholder to be used instead of settings.COMPRESS_URL during offline compression.
     # Affects manifest file contents only.
     URL_PLACEHOLDER = '/__compressor_url_placeholder__/'
+    # enables the output of Django CSP Nonce in script and style tags
+    COMPRESS_CSP_NONCE = False
 
     # Returns the Jinja2 environment to use in offline compression.
     def JINJA2_GET_ENVIRONMENT():
--- a/compressor/templates/compressor/css_file.html
+++ b/compressor/templates/compressor/css_file.html
@@ -1,1 +1,1 @@
-<link {% if request.csp_nonce %}nonce="_request_csp_nonce" {% endif %}rel="stylesheet" href="{{ compressed.url }}" type="text/css"{% if compressed.media %} media="{{ compressed.media }}"{% endif %}>
+<link {% if compress_csp_nonce %}nonce="_request_csp_nonce" {% endif %}rel="stylesheet" href="{{ compressed.url }}" type="text/css"{% if compressed.media %} media="{{ compressed.media }}"{% endif %}>
--- a/compressor/templates/compressor/css_file.html
+++ b/compressor/templates/compressor/css_file.html
@@ -1,1 +1,1 @@
-<link rel="stylesheet" href="{{ compressed.url }}" type="text/css"{% if compressed.media %} media="{{ compressed.media }}"{% endif %}>
+<link {% if request.csp_nonce %}nonce="_request_csp_nonce" {% endif %}rel="stylesheet" href="{{ compressed.url }}" type="text/css"{% if compressed.media %} media="{{ compressed.media }}"{% endif %}>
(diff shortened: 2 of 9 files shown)


iamareebjamal:
@diox Added tests and confirmed working in online mode. Just working on somethings for offline mode. Can you please check if you agree with the approach.

About offline mode, there are 2 options:

1. Add settings.COMPRESS_CSP_NONCE_TAG - this is the easiest one and more configurable, just need to change 1 setting to turn on or off
2. Add `{% compress js csp_nonce %}` - this is more involved

iamareebjamal:
Confirmed to be working in offline mode

iamareebjamal:
@diox Ping?

diox:
Sorry for the delay, been busy with other stuff. I haven't tried this yet but it looks good on the surface. 

Regarding offline, I am not sure why you'd want a specific tag or setting. Actually I'm not even sure you need `COMPRESS_CSP_NONCE` setting at all. Is there a scenario where making it work automatically when there is a nonce in the original content would be a bad thing ? 

One improvement I think you should do is have a method in the `CompressorMixin` to get the nonce from a `context` (that would, in the default implementation, `return context['request'].csp_nonce`). That way it's possible to override the behavior by using a custom templatetag.

iamareebjamal:
I remember that it was needed in offline compression because at the time, context does not have a nonce. If there is an alternative, I'll implement it.

About the method in `CompressorMixin`, I'll add that

diox:
I think the setting can be replaced by detecting if there is a nonce in the original content or not.

iamareebjamal:
Hmm, how will we detect that? If I match `nonce=`, it'll create false posiitves because it can be in the content as well

iamareebjamal:
Also, the setting gives you the ability to not write nonce in each and every javascript/CSS and get nonce in the output as evident in the test

diox:
We already have an HTML parser involved when compressing, so we can extract the necessary info in a reliable way and pass it down from there.

9mido:
Any ideas when this will be PR change will take effect? I am looking forward to this since my site uses django-csp.

iamareebjamal:
I haven't got enough time to finalize the requested changes after last review, however we are already using it in a production app without any issues. I will still see if I can get some time to change the implementation
However, note that only implementation detail is needed to change, everything else works fine

manickamx2:
@iamareebjamal @diox Wanted to bump this, any ideas if/when this PR could be merged? Thanks.

diox:
The conflicts would need to be solved and my suggestion from https://github.com/django-compressor/django-compressor/pull/989#issuecomment-593358003 implemented before we could consider merging this.

uniuuu:
Hi @diox 
Any quick workaround available before the PR will finalized?

diox:
The logic from that PR could be re-used in a custom class and templates inheriting from compressor. Besides that I don't know. You're welcome to try to pick that PR up again and re-submit it with my comments addressed (crediting original submitter).

Issue #979, opened 2019-12-20
`Extends` tag offline processing fails on relative paths


kdickerson:
Django 1.10 introduced relative paths for the `extends` tag, for example `{% extends '../base.html' %}`.  These fail to resolve during offline compression and `compress` catches a `TemplateDoesNotExist` exception at line 201 of `management/commands/compress.py`.

The reported path appears to be correct except that it isn't written absolutely (note that the first path starts with `/` while the failing 2nd path does not):

```
# my_app/templates/my_app/_base.html
....

# my_app/templates/my_app/show.html
{% extends './_base.html' %}
....

# compress-command error logging
Error parsing template /home/kdickerson/Code/my_project/my_app/templates/my_app/show.html: home/kdickerson/Code/my_project/my_app/templates/my_app/_base.html
```

It works correctly if I replace the relative path with a relative-to-template-loader-root path: `{% extends 'my_app/_base.html' %}`

karyon:
thanks for the report! while i won't be able to fix this myself anytime soon, we take PRs :)

kdickerson:
Whoops, mis-clicked - sorry, kids running around being crazy!

kdickerson:
Here's what I've learned: Django's calculation of relative template paths is done in [loader_tags.py:construct_relative_path](https://github.com/django/django/blob/master/django/template/loader_tags.py#L220) which actually works off of a template `name` rather than a `path`.  The `name` is the relative-to-template-loader path to the template.  That is, it constructs a new relative-to-template-loader path based on the existing relative-to-template-loader path.  (This is the `template_name` property on the `Origin` instance within the `parser`.)

In my above example, I may have a `TemplateView` which sets `template_name = 'my_app/show.html'`.  When `construct_relative_path` is executed, the value of `current_template_name` is `my_app/show.html` (the value of `origin.template_name`) and not `/home/kdickerson/Code/my_project/my_app/templates/my_app/show.html` (the value of `origin.name`).

Django-Compressor, however, is looking for templates manually and finds them all by absolute filesystem path and doesn't know what their relative-to-template-loader paths are (so `origin.name` and `origin.template_name` are the same).  This happens in [compress.py:compress](https://github.com/django-compressor/django-compressor/blob/master/compressor/management/commands/compress.py#L106).  So when it reaches the extends tag in the template it sends the full filesystem path to `construct_relative_path` and it fails.

[...]

PR #995, opened 2020-03-11
#979 - Correctly handle relative paths in "extends" tags.

kdickerson:
For django-compressor/django-compressor#979.

Django 1.10 introduced relative paths for the extends tag, for example `{% extends '../base.html' %}`. These fail to resolve during offline compression and compress catches a `TemplateDoesNotExist` exception.

This commit fixes the issue by passing a relative path to `templates.update`.


karyon:
thanks! and congrats for getting through that process! :)

PR #1003, opened 2020-06-30
Create automatic publish to PyPi via GitHub actions

Andrew-Chen-Wang:
If you're busy, you may want to consider this method of publishing to PyPi just to get patches out of the way. Just create a PyPi token on the site for this package, put it in your GitHub secrets with the name PYPI_PASSWORD, and your all set.

That way, you can just create a release via the GitHub website in a couple clicks and be on your way.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,31 @@
+# This workflows will upload a Python Package using Twine when a release is created
+# For more information see: https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries
+
+name: Upload Python Package
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  deploy:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install setuptools wheel twine
+    - name: Build and publish
+      env:
+        TWINE_USERNAME: __token__
+        TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+      run: |
+        python setup.py sdist bdist_wheel
+        twine upload dist/*


karyon:
Sorry for the delay. This is pretty cool, thanks.

Why does it say __token__ as the username? The github template also puts the username into the secrets.

@diox I can't manage the pypi project, so i guess I cannot create a token. Could you do that? Or give me manage privileges (might be good anyway to increase the bus factor :) )

Andrew-Chen-Wang:
No prob @karyon ! Tokens are per package whereas inputting your username and password gives access to all your packages.

karyon:
Thanks! though I was rather referring to the syntax `__token__` instead of something like `${{ secrets.PYPI_TOKEN }}`. What's that variable and where does it come from?

Andrew-Chen-Wang:
Ah I'm pretty sure `__token__` is just a string in GitHub's yaml files (e.g. in [cachalot](https://github.com/noripyt/django-cachalot/blob/d0b5213014e80db2ecb78cbf0474e791414d53d8/.github/workflows/publish.yml#L27-L28), I've got it like that but in cookiecutter-django, I think I put it in double quotes, and it was also acceptable). Maybe the double quotes are simply necessary in case of spaces?

Maybe that's the case for environment variables like `${{ secrets.PYPI_TOKEN }}`? I'm not totally sure tbh.

karyon:
alright, thanks. 

@diox, could you create and save that token or give me permissions to do so?

diox:
I created the token and stored in in the repos secret under the `PYPI_PASSWORD` name.

tbh while it's nice to see this being automated, this was never the part that was blocking making new releases. The more annoying bit is coming up with a nice changelog - we've looked into automating this in the past and while it helps, there is always some manual work involved to make it a little more coherent and elegant. And of course you still need to bump the version in `compressor/__init__.py`, create and then push the corresponding tag.

diox:
BTW, it would be nice to double down and replace Travis with Github Actions, with the future of Travis for open-source repos looking so grim...

Andrew-Chen-Wang:
@diox Not sure if what cookiecutter-django's doing is overkill for changelog (uses GitHub actions and just checks by list of PRs). Another option could be: https://github.com/Michael-F-Bryan/auto-changelog.

When it comes to actually getting this to auto commit, take a look at this GitHub action workflow: 

```
name: "Dependabot Automerge - Action"
on:
  pull_request:
jobs:
  worker:
    runs-on: ubuntu-latest
if: github.actor == 'dependabot[bot]'
    steps:
      - name: automerge
        uses: actions/github-script@0.2.0
        with:
          script: |
            github.pullRequests.createReview({
              owner: context.payload.repository.owner.login,
              repo: context.payload.repository.name,
              pull_number: context.payload.pull_request.number,
              event: 'APPROVE'
            })
            github.pullRequests.merge({
              owner: context.payload.repository.owner.login,
              repo: context.payload.repository.name,
              pull_number: context.payload.pull_request.number
            })
          github-token: ${{github.token}}
```
This was from https://handsonappsec.medium.com/automerge-github-dependabot-alerts-with-github-actions-7cd6f5763750

Obviously you can remove that last part (merge).

PR #1006, opened 2020-07-16
add support for type=module & attr nomodule

dmlb:
Add support for the attribute 'nomodule' on `<script>` tags inside `JsCompressor`
Add support for  output mode 'module' for script tag with `<script type="module">` in `base.py` and supporting HTML Template files
Fix error message for 'preload' output mode and add new 'module' output mode
Add test for module output mode
Add to attributes test for nomodule attribute

potentially closes:
https://github.com/django-compressor/django-compressor/issues/1001

### DIFF
--- a/docs/usage.txt
+++ b/docs/usage.txt
@@ -6,7 +6,7 @@ Usage
 .. code-block:: django
 
     {% load compress %}
-    {% compress <js/css> [<file/inline/preload> [block_name]] %}
+    {% compress <js/css> [<file/inline/preload/nomodule> [block_name]] %}
     <html of inline or linked JS/CSS>
     {% endcompress %}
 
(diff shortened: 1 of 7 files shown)


docs/usage.txt
```diff
@@ -63,6 +63,33 @@ Result:
         <link rel="preload" href="/static/CACHE/js/d01466eb4fc6.js" as="script" />
 
 
+
+ES5 nomodule & ES6 type="module"
+
+Adding the ``nomodule`` attribute will generate be passed attribute to the script tag for the compressed resource in the template.
+Adding the ``module`` parameter will generate the script tag with type="module" for the compressed resource in the template:
```
karyon, line 70: I'd suggest clarifying that `nomodule` needs to go into the script html tag, and `module` needs to go into the compress templatetag.

docs/usage.txt
```diff
@@ -63,6 +63,33 @@ Result:
...
+
+    .. code-block:: django
+
+    {% compress js %}
+        <!-- ES5 must be in separate compress (split types) -->
+        <script nomodule async src="static/js/es5.js"></script>
+    {% endcompress %}
+
+    {% compress js module %}
```
karyon, line 79: I think it would be better if this supported the `type="module"` syntax instead of the additional templatetag parameter. Then i would also worry less if it still worked properly with COMPRESS_ENABLED = False.

karyon (commented):
Sorry for the long delay.


Since we modify the filename, and the filename is the modulename, the compressed module can't be imported by other modules anymore (which might be fine depending on the use case). I think this should be mentioned in the documentation.

Also, modules can't be concatenated, so this option will break if a user puts a second file into the  really makes sense only for a single file in the compress tag. This should also be documented. A fix might be to disable concatenation if any of the contained files is a module, but I'd rather keep it simple.

So, I appreciate this feature to enable a specific limited use case, but it should be documented that this is really just a specific limited use case, and no support for modules in general.

karyon:
@diox do you have any comment on module support in general?

diox:
I am not sure it's worth doing. Modules are difficult to support properly for compressor. As you pointed out, the filename needs to be known in advance and you likely don't want to use compressor on them in the first place - you probably don't want concatenating the scripts either. I just don't see the point of adding complexity for it in compressor.

docs/usage.txt
```diff
@@ -6,7 +6,7 @@ Usage
 .. code-block:: django
 
     {% load compress %}
-    {% compress <js/css> [<file/inline/preload> [block_name]] %}
+    {% compress <js/css> [<file/inline/preload/nomodule> [block_name]] %}
```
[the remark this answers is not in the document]
dmlb: `nomodule` is an attribute itself, but `module` is a `type` attribute option - they have a relationship but they are not the same
this is for how the library strips all attributes

dmlb:
> I am not sure it's worth doing. Modules are difficult to support properly for compressor. As you pointed out, the filename needs to be known in advance and you likely don't want to use compressor on them in the first place - you probably don't want concatenating the scripts either. I just don't see the point of adding complexity for it in compressor.

yes, it is for very small specific use case and was not able to solve for proper module splitting :)

PR #1010, opened 2020-08-22
extra logging in template tag

skiv23:
Pass a logger instance from templatetag to Compressor in order to add extra info for debugging.


karyon:
thanks!

PR #1024, opened 2020-11-16
Add a management command option to exclude files ...

### LINKED ISSUE #785: Offline compress command add `--ignore=PATTERN` support
Exception may occur when angular template file locate in template directory. That will be nice if `--ignore` parameter is supported like `collectstatic` command.

```
-i PATTERN, --ignore=PATTERN
                      Ignore files or directories matching this glob-style
                      pattern. Use multiple times to ignore more.
```

Then, we can filter some frontend template file like `.ng.html`.

spapas:
from offline compression. It uses the fnmatch module to
match the files and will output the list of excluded
templates if verbose >=2.

Example call:

`python manage.py compress -x "forms/*.html" -x "test*/*.html"`

This shoud fix #785

### DIFF
--- a/compressor/management/commands/compress.py
+++ b/compressor/management/commands/compress.py
@@ -42,6 +42,9 @@ class Command(BaseCommand):
                                  "supported. It may be a specified more than once for "
                                  "multiple engines. If not specified, django engine is used.",
                             dest="engines")
+        parser.add_argument('--exclude', '-x', default=[], action="append", dest='exclusions',
+                            help="Add excluded files using a glob patter. It can be used "
+                                 "multiple times to add more exclusions." )
 
     def get_loaders(self):
         template_source_loaders = []
@@ -86,7 +89,7 @@ class Command(BaseCommand):
 
         return parser
 
-    def compress(self, engine, extensions, verbosity, follow_links, log):
+    def compress(self, engine, exclusions, extensions, verbosity, follow_links, log):
         """
         Searches templates containing 'compress' nodes and compresses them
         "offline" -- outside of the request/response cycle.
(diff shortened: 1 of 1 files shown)


diox:
Hi,

Please include tests for the changes introduced in this PR.

spapas:
Hello @diox, I checked the tests but the `compress` management command isn't tested that thoroughly... Could you point me to the right direction?

diox:
`test_offline.py` should have decent (though far from perfect I agree) coverage. It uses the `test_templates` and `test_templates_jinja2` directories.

raphaelm:
Any way we can help moving this forward?

diox:
Hi,

My comment from https://github.com/django-compressor/django-compressor/pull/1024#issuecomment-728123258 still applies. In order for it to be considered, pull request needs to come with tests and docs for the changes added. 

So if you want for this feature to be added, try forking the original poster repos (or work directly with them) and add tests & docs.

raphaelm:
Hi diox, thanks.

We decided it's not the right solution for us anyways, so I'm not able to invest the time at the moment. With is solution, we'd need to train every user of our software to pass the new command-line parameters, but we'd rather like a solution that "just works" when you say "manage.py compress", so we're now using a custom django-level template loader that excludes the 3rd-party templates we don't actually use.

spapas:
@raphaelm could you clarify a bit more on your solution? Which is the template loader you use? It may be helpful for me also.

Thank you

raphaelm:
Yeah, sure, you can have a look at our work in progress here: https://github.com/pretix/pretix/pull/2459/files

rvanlaar:
Here's our solution.

Place file in `management/commands` directory as compress_vendor.
Replace `APPLICATION_DIR` with the root directory of your application.
Run it with `./manage.py compress_vendor`

```python
from __future__ import annotations

from pathlib import Path
from typing import cast

from compressor.management.commands import compress
from django.template.backends.django import DjangoTemplates
from django.template.loaders.app_directories import Loader


class CompressLoader(Loader):
    """Loader which only returns template paths inside this application."""

    def get_dirs(self) -> list[str | Path]:
        ds = super().get_dirs()
        return [d for d in ds if "APPLICATION_DIR" in str(d)]


class Command(compress.Command):
    """Compressor for only edubroker_kom templates."""

    def get_loaders(self) -> list[Loader]:
        template = cast(DjangoTemplates, engines["django"])
        return [CompressLoader(template.engine)]
```

PR #1025, opened 2020-11-17
added compress_options to sekizai postprocessor

CaptainChaos:
Hi guys,
after I cancelled my last PR due to missing features, I now added an improved way to configure the sekizai postprocessor. Specifically, I missed a way to

- compress custom-named blocks, since the old postprocessor only supported the names "js" and "css"
- generate preload or inline compress blocks

My idea was that a sekizai block can now contain an html comment &lt;!-- compress_options [params] --&gt;, where [params] can be

- one of "js" or "css"
- one or multiple of "file", "preload" and "inline"

If the comment is missing, the behavior of the postprocessor doesn't change. This way, I can for example include &lt;!-- compress_options css file preload --&gt; in my css sekizai block (with {% addtoblock "css" %}), and the postprocessor generates a compressed file and also the correct preload link.

If I should change anything, let me know.

Best regards!

### DIFF
--- a/compressor/contrib/sekizai.py
+++ b/compressor/contrib/sekizai.py
@@ -12,6 +12,7 @@ from compressor.conf import settings
 from compressor.utils import get_class
 
 from django.template.base import TextNode
+import re
 
 
 def compress(context, data, name):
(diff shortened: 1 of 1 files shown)


CaptainChaos:
I added a defer option for scripts, because sometimes this makes sense for page speed optimizations

karyon:
Hi! sorry for the late reply. is this params-in-html-comments a thing in sekizai? if yes that's fine, otherwise it appears too hacky to me to include it. Unfortunately I don't know sekizai at all so I can't give any design recommendations.

PR #1030, opened 2020-12-03
Fix cache key type

josephtyler:


karyon:
Hi, what problem were you trying to solve?

Issue #985, opened 2020-01-02
Deprecate COMPRESS_*_FILTERS settings


karyon:
They were replaced by COMPRESS_FILTERS in #900, i would simplify that to only have one version of those settings.

karyon:
also, we could remove references to JSMinFilter and CSSMinFilter, and only use rJSMinFilter and rCSSMinFilter instead.

diox:
We announced in the changelog that `COMPRESS_CSS_FILTERS` and `COMPRESS_JS_FILTERS` were deprecated, but they aren't deprecated in the code, and also the docs encourage you to use `COMPRESS_FILTERS` instead, it doesn't say they are deprecated either.

So I think the first step would be to update raise a proper deprecation warning, update the docs, and then remove them in a future release.

Issue #888, opened 2017-11-08
JS files are minified by default, but CSS files are not


badbye:
Hi, I am using Django(1.8.2) and django-compressor(2.2). The compressed js works good, however, the css file is not compressed.

Here is part of my template:
```html
{% load staticfiles %}
{% load compress %}
{% compress css %}
<link rel="stylesheet" href="{% static 'landing/style.css' %}" type="text/css"></head>
{% endcompress %}
{% compress js %}
<script src="{% static 'landing/js/jquery.min.js' %}"></script>
<script src="{% static 'landing/js/ads.js' %}"></script>
<script src="{% static 'landing/js/ads.js' %}"></script>
{% endcompress %}
```
It do generate a css file, but it is exactly the same with the original file.

karyon:
please post your settings.py.

badbye:
```python
INSTALLED_APPS = (
    'django.contrib.admin',
    'django.contrib.auth',
    'django.contrib.contenttypes',
    'django.contrib.sessions',
    'django.contrib.messages',
    'django.contrib.staticfiles',
    'compressor',
)

STATIC_URL = '/static/'

STATICFILES_DIRS = [
    os.path.join(BASE_DIR, "static"),
]

STATICFILES_FINDERS = (
  'django.contrib.staticfiles.finders.FileSystemFinder',
  'django.contrib.staticfiles.finders.AppDirectoriesFinder',
  'compressor.finders.CompressorFinder',
)

COMPRESS_ENABLED = True
COMPRESS_OUTPUT_DIR = 'compress'
STATIC_ROOT = '/home/service/growth/static_file/static'
```

karyon:
css is not compressed by default, see the default setting here: https://django-compressor.readthedocs.io/en/latest/settings/#django.conf.settings.COMPRESS_CSS_FILTERS. i admit it is weird that this is done for js by default and not for css. but we probably won't change that unless we do some release with other major breaking changes.

badbye:
Sorry, I am not familiar with CSS. Should I set the `COMPRESS_CSS_FILTERS` variable to make it work for css? Which option should I use?

badbye:
I see. This line works:
```
COMPRESS_CSS_FILTERS = ['compressor.filters.css_default.CssAbsoluteFilter', 'compressor.filters.cssmin.rCSSMinFilter']
```

Thank you.

karyon:
i'll leave this open so we can discuss this in case we make such a release.

[...]

PR #1043, opened 2021-02-28
Move rcssmin to extras_require

tufedtm:
What about moving rjsmin and other packages to the extras_require?

### DIFF
--- a/docs/behind-the-scenes.txt
+++ b/docs/behind-the-scenes.txt
@@ -9,7 +9,7 @@ The goal is to explain what the main template tag, {% compress %}, does
 behind the scenes, to help you debug performance problems for instance.
 
 Offline compression
--------------
+-------------------
 
 If offline compression is activated, the {% compress %} tag will try to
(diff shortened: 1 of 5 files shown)


browniebroke:
Nice one. It would be nice for end users to not have packages installed if they don't need them...

diox:
In https://github.com/django-compressor/django-compressor/issues/888 and https://github.com/django-compressor/django-compressor/issues/985 we kinda discussed moving the opposite direction:
- Depend on both `rjsmin` and `rcssmin`
- Enable `rJSMinFilter` and `rCSSMinFilter` by default

The idea would be to have a fully functional compressor for both css and js out of the box - which is what most users want.

diox:
Wontfixing because I think the approach of providing out of the box css and js compression is more desirable for something like compressor (compared to modern assets management tools, its main strength is the fact that it can just work out of the box with 0 tweaks, even if it's strongly suggested for advanced users to customize the settings)

Issue #1044, opened 2021-03-19
Defer non-critical CSS (preload)


brunoprograma:
Well, in my case the "preload" didn't even work for CSS files. Browsers don't even load them. But, even for those who are using and working, I believe there is a lack of optimization for an asynchronous load, according to the documentation available at:

https://web.dev/defer-non-critical-css/

The program is creating only new attributes in a <link> tag (rel="preload" and as="style"), but the correct thing would be to make it look like the example below (taken from the previous link):

`<link rel="preload" href="styles.css" as="style" onload="this.onload=null;this.rel='stylesheet'">`
`<noscript><link rel="stylesheet" href="styles.css"></noscript>`

Actually, web.dev is a great source of information to any new feature in this project.

Check #1045

diox:
See https://github.com/django-compressor/django-compressor/pull/1045#issuecomment-803134423

PR #1045, opened 2021-03-19
Defer non-critical CSS

brunoprograma:
Defer non-critical CSS

### DIFF
--- a/compressor/templates/compressor/css_preload.html
+++ b/compressor/templates/compressor/css_preload.html
@@ -1,1 +1,2 @@
-<link rel="preload" href="{{ compressed.url }}" as="style" />
+<link rel="preload" href="{{ compressed.url }}" as="style" onload="this.onload=null;this.rel='stylesheet'">
+<noscript><link rel="stylesheet" href="{{ compressed.url }}"></noscript>


jsumnerPhD:
Just a quick question: is this ok for csp with the inline script in the online attribute?

brunoprograma:
I honestly don't know anything about this policy. The suggestion was in the role of developer, following the Google Developers usage guidance given in the Lighthouse report. Check #1044

brunoprograma:
> Just a quick question: is this ok for csp with the inline script in the onload attribute?

I honestly don't know anything about this policy. The suggestion was made following the Google Developers usage guidance given in the Lighthouse report. Check #1044

diox:
> Just a quick question: is this ok for csp with the inline script in the onload attribute?

It's not.

diox:
Hi,

Thanks for the PR, but I'm going to reject it:
- This was opened before we got to discuss whether or not it's a good idea in the django-compressor app itself (I don't think it is, because it introduces inline script)
- Tests are not passing
- There should be docs, because it's not necessarily the expected behavior.

You're welcome to customize that template in your own project if you want to preload css this way.

PR #1047, opened 2021-04-12
Fix coverage reporting

browniebroke:
I noticed in recent GH action build logs that [codecov couldn't find the coverage data](https://github.com/django-compressor/django-compressor/runs/2318689313?check_suite_focus=true#step:6:37). thought I broke this with #1046 but it seems like it was broken already (last data is a year old?).

When I ran it again with the `codecov` python lib, there was another error that seemed to indicate we should update coverage to v5:

> Error running `['/opt/hostedtoolcache/Python/3.5.10/x64/bin/python', '-m', 'coverage', 'xml', '-i']`: returncode=1, output=b"Couldn't use data file '/home/runner/work/django-compressor/django-compressor/.coverage': Looks like a coverage 4.x data file. Are you mixing versions of coverage?\n"

From my testing, it seems like it should do the trick.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,4 +46,6 @@ jobs:
           pip install tox
           python setup.py install_egg_info
       - run: tox -e ${{ matrix.versions.toxenv }}
-      - uses: codecov/codecov-action@v1.3.2
+      - run: |
+          pip install codecov
+          codecov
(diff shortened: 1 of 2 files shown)


diox:
Thanks! The test requirements would benefit from being kept up to date more closely... TBH I haven't spent much time on compressor itself for quite a while...

Issue #984, opened 2020-01-02
Drop django 1.11 support 


karyon:
As per the discussion in https://github.com/django-compressor/django-compressor/pull/965

Cabalist:
I would love to see the current state of the code be released.   :)

diox:
Fixed by https://github.com/django-compressor/django-compressor/pull/1048

There will be a new release shortly.

PR #1048, opened 2021-04-12
Update Django/Python support

browniebroke:
- Add Python 3.9
- Add Django 3.1 and 3.2

However, to achieve that, we also need to:

- Bump django-sekizai to v2 (earlier versions don't support Django 3.1)
- Drop support for Django 1.11 ([django-sekizai 2 requires Django 2.2 or higher](https://github.com/django-cms/django-sekizai/blob/e12d8159bc1769a36472408e5efe97bbf65de9f9/setup.py#L8))

According to [this page](https://www.djangoproject.com/download/), we've passed the end of extended support for Django 1.11 by more than a year (April 1, 2020), so I think it should be ok to make this breaking change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,14 +14,6 @@ jobs:
       matrix:
         versions:
           - python: 3.5
-            toxenv: py35-1.11.X
-          - python: 3.6
-            toxenv: py36-1.11.X
-          - python: 3.7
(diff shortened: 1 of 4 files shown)


diox:
Agreed. That fixes https://github.com/django-compressor/django-compressor/issues/984

diox:
Thanks for the patch!

PR #1055, opened 2021-04-21
Fix changelog entry for deprecation warning

browniebroke:
I think the changelog entry for the latest release currently lists the new unified setting as deprecated, which is probably not what was intended 😄


diox (approved):
:facepalm: thanks

PR #1062, opened 2021-06-09
Update classifiers

jaap3:
As far as I can tell django-compressor isn't tested on 1.11 anymore, but it is tested on Django 3.2


jaap3:
Not sure why the tests are failing on Python 3.9, this seems unrelated to this PR

jaap3:
@diox could you re-run the tests? The failure has noting to do with this PR

diox:
Thanks, we can ignore the failures for this change

PR #1063, opened 2021-06-09
Remove unused django-discover-runner dependency

jaap3:


jaap3:
@diox could you re-run the tests? The failure has noting to do with this PR

PR #1081, opened 2021-11-15
update rjsmin to version 1.2.0

BoPeng:
rjsmin is updated to 1.2.0 to [fix a bug that incorrectly compresses some (mostly minified) `.js` files](https://github.com/ndparker/rjsmin/issues/17).


BoPeng:
saw another PR.

PR #1085, opened 2021-11-23
Update RCSSMIN version :)

Dhruvacube:
The old version has problems in installing but not the new version of rcssmin, so I updated its version number


carltongibson (approved):
Super. Thanks for this! 🥇

PR #1086, opened 2021-11-26
Fix offline manifest race condition

cuu508:
django-compressor avoids rendering the same node multiple times by checking for a key in offline_manifest:

```
    if key in offline_manifest:
        continue

    [... render template ...]

    offline_manifest[key] = result
```

Multiple _compress_template calls can run concurrently, creating a race condition: each checks for a key in offline manifest, then each proceeds to render template. Finally they try to save the same file at the same time, causing #1082.

My proposed fix is to atomically
 * check if the manifest key exists
 * if it doesn't exist, set it to a placeholder value (None)

So, in nutshell, the first "if" part becomes:

```
    with offline_manifest_lock:
        if key in offline_manifest:
            continue

        offline_manifest[key] = None
```

I'm not sure where to store the lock, I've put it at the module level currently. Perhaps there's a way to avoid the lock entirely.

### DIFF
--- a/compressor/management/commands/compress.py
+++ b/compressor/management/commands/compress.py
@@ -2,6 +2,7 @@
 import os
 import sys
 import concurrent.futures
+from threading import Lock
 
 from collections import OrderedDict, defaultdict
 from fnmatch import fnmatch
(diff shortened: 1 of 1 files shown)


carltongibson (approved):
Hi @cuu508 — thanks for this. 

> Perhaps there's a way to avoid the lock entirely.

It looks OK to me. Maybe there's another approach but this seems simple enough to go with. 

You're officially Hero of the Month here 🥇

PR #1089, opened 2021-12-13
Create .readthedocs.yaml

jezdez:
Congrats on the big 3.0! I noticed the docs weren't building with an Python 2.7 backend error. I've fixed it by configuring it to the Python 3 backend. I've also enabled running doc builds automatically on pull-requests.

This PR helps making configuration in case of build errors easier going forward.

Docs: https://docs.readthedocs.io/en/stable/config-file/v2.html

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,31 @@
+# .readthedocs.yaml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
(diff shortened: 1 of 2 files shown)


jezdez:
Ah, huh, it seems as if the webhook is somehow not correctly set up and it can't create the checks for the pull request, it should have linked to https://django-compressor--1089.org.readthedocs.build/en/1089/ automatically. I just added the "checks run" event to the webhook as well, maybe that helps?

carltongibson (approved):
This looks right. Thanks @jezdez!

PR #1090, opened 2021-12-14
Update conf.py

idiosincratica:
configure_precompilers didn't through an exception due to missing commas in COMPRESS_PRECOMPILERS configuration tuple because isinstance(value, (list, tuple)) checked inner tuple that was ok


carltongibson (commented):
Hi @idiosincratica — could you add a regression test for this please? Thanks!

Issue #1099, opened 2022-02-11
Intermittent failure in offline compression


jtremesay-sereema:
Hi,

We use `django-compressor` (v3.1) in our Django (v4.0.2) application. Our CI randomly fails (something like 1 run in 20) on the compress step with errors like this one:

```
#18 [16/17] RUN python manage.py compress --extension .html,.xsl
#18 3.681 CommandError: An error occurred during rendering registration/password_change_form.html: [Errno 2] No such file or directory: '/code/sereema/staticfiles/CACHE/js/output.5840362cca8f.js'
```

If I read correctly, this is a known issue (#1082) that was fixed in v3.0. But it seems like the fix was not enough.

jtremesay-sereema:
I can try

jtremesay-sereema:
I have created a dummy django project that trigger the bug most of the time

https://github.com/jtremesay-sereema/django-compressor-issue-1099

```
$ pip install -r requirements.txt
$ ./manage.py collectstatic --no-input
$ ./manage.py compress
CommandError: An error occurred during rendering foo/gen/foo_669.html: [Errno 2] No such file or directory: '/Users/jtremesay/projects/compressor_issue/staticfiles/CACHE/js/output.6169dc6ebd1b.js'
Compressing...
$ ./manage.py compress
CommandError: An error occurred during rendering foo/gen/foo_931.html: [Errno 2] No such file or directory: '/Users/jtremesay/projects/compressor_issue/staticfiles/CACHE/js/output.6169dc6ebd1b.js'
Compressing...
$ ./manage.py compress
CommandError: An error occurred during rendering foo/gen/foo_164.html: [Errno 2] No such file or directory: '/Users/jtremesay/projects/compressor_issue/staticfiles/CACHE/js/output.6169dc6ebd1b.js'
Compressing...
$ ./manage.py compress
CommandError: An error occurred during rendering foo/gen/foo_3.html: [Errno 2] No such file or directory: '/Users/jtremesay/projects/compressor_issue/staticfiles/CACHE/js/output.6169dc6ebd1b.js'
Compressing...
$ ./manage.py compress
CommandError: An error occurred during rendering foo/gen/foo_433.html: [Errno 2] No such file or directory: '/Users/jtremesay/projects/compressor_issue/staticfiles/CACHE/js/output.6169dc6ebd1b.js'
Compressing...
$ ./manage.py compress
[...]

PR #1100, opened 2022-02-22
Fixed compressor.base.Compressor to avoid duplicate save() calls

cuu508:
This is a proposed fix for #1099.

### DIFF
--- a/compressor/base.py
+++ b/compressor/base.py
@@ -1,6 +1,7 @@
 import os
 import codecs
 from importlib import import_module
+from threading import Lock
 from urllib.request import url2pathname
 
 from django.core.files.base import ContentFile
@@ -29,6 +30,8 @@ class Compressor:
     """
 
     output_mimetypes = {}
+    seen_filepaths = set()
+    seen_filepaths_lock = Lock()
 
     def __init__(self, resource_kind, content=None, output_prefix=None,
                  context=None, filters=None, log=None, verbosity=1, *args, **kwargs):
@@ -346,8 +349,22 @@ class Compressor:
         the appropriate template with the file's URL.
         """
         new_filepath = self.get_filepath(content, basename=basename)
-        if not self.storage.exists(new_filepath) or forced:
+        if forced:
+            # forced=True means this is offline generation (manage.py compress).
+            # We should overwrite all existing files.
+            # But, to avoid overwriting a file more than once in a single run,
+            # keep track of already handled filepaths in `seen_filepaths` class
+            # variable. If a filepath is already there, *don't* call
+            # self.storage.save() for it.
+            with self.seen_filepaths_lock:
+                skip_save = new_filepath in self.seen_filepaths
+                self.seen_filepaths.add(new_filepath)
+        else:
+            skip_save = self.storage.exists(new_filepath)
+
+        if not skip_save:
             self.storage.save(new_filepath, ContentFile(content.encode(self.charset)))
+
         url = mark_safe(self.storage.url(new_filepath))
         return self.render_output(mode, {"url": url})
 
--- a/compressor/tests/test_offline.py
+++ b/compressor/tests/test_offline.py
@@ -12,6 +12,7 @@ from django.template import Context, Origin, Template
 from django.test import override_settings, TestCase
 from django.urls import get_script_prefix, set_script_prefix
 
+from compressor.base import Compressor
 from compressor.cache import flush_offline_manifest, get_offline_manifest
 from compressor.exceptions import OfflineGenerationError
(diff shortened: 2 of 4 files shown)


diox:
Unrelated test failures: https://github.com/django-compressor/django-compressor/issues/1101

diox:
Updating your branch with `develop` should fix the slimit related test failures.

cuu508:
Done!

diox:
So, this works, so if we don't find anything better, I'll merge this. 

I don't love the fact that we're handling all this in the `Compressor` class, though - this kind of logic feels out of place here. 

As you mentioned, the call chain up to the `compress` command is long, so it would be a pain to modify, but I wonder if it's not the better path. Any thoughts on going that route ?

cuu508:
> I don't love the fact that we're handling all this in the Compressor class, though - this kind of logic feels out of place here.

A couple ideas:

A. In a `manage.py compress` run, don't run `storage.save(...)` in the Compressor class. Instead, pass the rendered result up the call chain, and make it the responsibility of the management command to run `storage.save(...)`. This would be a messy change, would need to pass the rendered content up the call chain. Many small changes and potential for new bugs.

B. Make a StorageProxy class which wraps the normal storage class, but filters out the duplicate `save()` calls. Make the management command use this class instead of the normal storage directly. This adds an extra level of indirection – and there are quite few already! Would become harder to understand the code, figure out how everything interacts.

> As you mentioned, the call chain up to the compress command is long, so it would be a pain to modify, but I wonder if it's not the better path. Any thoughts on going that route ?

Yep, we can try that and see how it looks. (A wild idea I had was to piggy-back on the existing `forced` variable – instead of passing True, pass a set of filenames in it. But that's a giant hack)

diox:
> A. In a manage.py compress run, don't run `storage.save(...)` in the Compressor class. Instead, pass the rendered result up the call chain, and make it the responsibility of the management command to run `storage.save(...)`. This would be a messy change, would need to pass the rendered content up the call chain. Many small changes and potential new bugs.

Yeah. The `Compressor` class already has the ability to render in different modes (*) so some of the bits we need are already there.

One could argue that the whole `forced` variable mechanism is already messy... Though most of the bugs related to it have been squashed over the years... Really, the whole concept of offline generation is a bit crazy when you think about it for too long :-)

(*) notably, it has an inline mode that doesn't save to a file, but it's probably best to create a new one dedicated to offline generation, to be explicit.

> B. Make a StorageProxy class which wraps the normal storage class, but filters out the duplicate `save()` calls. Make the management command use this class instead of the normal storage directly. This adds an extra level of indirection – and there are quite few already! It would become harder to understand the code, and figure out how everything interacts.

Interesting idea. My first instinct was be to reject it because of the extra indirection as you said, but it's worth noting that compressor already has a [custom storage class](https://github.com/django-compressor/django-compressor/blob/be782489af21b51cad2e8a60dd6b0a3bed20bff3/compressor/storage.py#L12) anyway. In fact, I wonder if it's not part of the problem, since its `get_available_name()` deletes the file if it exists before writing it to disk ! This is done to bypass django's duplicate check, since by default it doesn't overwrite existing files and chooses a new name instead, and we really care about the name. But the fact that it's deleting the file first looks an awful lot like the kind of problem we're dealing with here...

> Yep, we can try that and see how it looks. (A wild idea I had was to piggy-back on the existing forced variable – instead of passing True, pass a set of filenames in it. But that's a giant hack)

Yeah I had the same thought initially but... That's not really better than the class variable.

diox:
Digging into the storage class and `get_available_name`: django has this:
https://github.com/django/django/blob/e0442a628eb480eac6a7888aed5a86f83499e299/django/core/files/storage.py#L300-L336

This is why our `get_available_name()` deletes first. Trying to force the file to be overwritten would lead to other issues - in fact, as highlighted by the comment in the code linked above, django purposely tries to avoid that situation to prevent two threads from writing to the same file at the same time.

But still, we control the process responsible for the writes here, so we could just skip the write through the lock + seen filenames mechanism. 

Interestingly, the whole issue might also potentially exist for compressor non-offline mode, and fixing this at the storage class might fix it for that use-case as well. Not sure we need to care about that too much though.

cuu508:
On putting the seen filenames in `CompressorFileStorage` – that would be a great place to put it.

The storage class is configurable via `settings.COMPRESS_STORAGE` – and in theory the end user can configure compressor to use a storage class that is not a subclass of `CompressorFileStorage`. Should we check for that, and emit a warning?

diox:
A warning sounds like a good idea, yeah.

diox:
https://github.com/django-compressor/django-compressor/pull/1107 addressed this differently.

Issue #1103, opened 2022-03-01
Non-deterministic failures when running with pytest-xdist


spookylukey:
When running Django tests with pytest and pytest-xdist, I get occasional failures.

This extension runs tests in parallel, in multiple processes (which is very helpful due to large speedups) . Sometimes, but not always, it seems we get interference between the runs, and the following stacktrace:
```
/home/luke/.virtualenvs/rpr/lib/python3.9/site-packages/compressor/base.py:245: in hunks
    yield self.handle_output(kind, value, forced=True,
/home/luke/.virtualenvs/rpr/lib/python3.9/site-packages/compressor/base.py:338: in handle_output
    return output_func(mode, content, forced, basename)
/home/luke/.virtualenvs/rpr/lib/python3.9/site-packages/compressor/base.py:350: in output_file
    self.storage.save(new_filepath, ContentFile(content.encode(self.charset)))
/home/luke/.virtualenvs/rpr/lib/python3.9/site-packages/django/core/files/storage.py:54: in save
    return self._save(name, content)
/home/luke/.virtualenvs/rpr/lib/python3.9/site-packages/django/core/files/storage.py:298: in _save
    os.chmod(full_path, self.file_permissions_mode)
```

Resulting in:
```
FileNotFoundError: [Errno 2] No such file or directory: '/home/luke/.../static/CACHE/css/main.aaa6e4bfc1a0.css'
```
The line it is failing on:
```python
    
        if self.file_permissions_mode is not None:
            os.chmod(full_path, self.file_permissions_mode)
```

This looks very similar to  #1099 but I'm putting it in a different issue, because this seems more likely to be related to multi-process race conditions.

diox:
I'm pretty sure this is the same underlying problem as https://github.com/django-compressor/django-compressor/issues/1099.

I believe that ultimately the problem is a race condition cause by the fact that compressor really wants to guarantee what the filename is going to be when writing files. Because django doesn't let it do that, it tries to delete the file first. Unfortunately multiple processes (or threads in #1099) are trying to write to the same file at the same time, and we end up with this race condition where sometimes the file is no longer present...

We're discussing a fix in https://github.com/django-compressor/django-compressor/pull/1100 but it's difficult without completely refactoring how compressor deals with files. We can avoid the problem for a single `compress` process using multiple threads, but it's going to be more difficult in the general case like this one which involves multiple processses.

diox:
[...]

PR #1105, opened 2022-03-05
Implemented NoDuplicatesStorageProxy, for use in "compress" command

### LINKED ISSUE #1099: Intermittent failure in offline compression

cuu508:
This is my second attempt to fix #1099.

I added a `NoDuplicatesStorageProxy`, which delegates all method calls to DefaultStorage, *except* it filters out duplicate `save()` calls.

I updated the `compress` management command to use `NoDuplicatesStorageProxy` by monkey-patching `compressor.storage.default_storage`.

The PR includes a test case which exercises the changes.

### DIFF
--- a/compressor/management/commands/compress.py
+++ b/compressor/management/commands/compress.py
@@ -16,6 +16,8 @@ from django.utils.encoding import smart_str
 from django.template.loader import get_template  # noqa Leave this in to preload template locations
 from django.template import engines
 
+
+from compressor import storage
 from compressor.cache import get_offline_hexdigest, write_offline_manifest, get_offline_manifest
 from compressor.conf import settings
 from compressor.exceptions import (OfflineGenerationError, TemplateSyntaxError,
@@ -191,6 +193,9 @@ class Command(BaseCommand):
         nodes_count = 0
         offline_manifest = OrderedDict()
         errors = []
+        # Replace compressor.storage.default_storage with a proxy object
+        # which will filter out duplicate save() calls
+        storage.default_storage = storage.NoDuplicatesStorageProxy()
 
         for context_dict in contexts:
             compressor_nodes = OrderedDict()
--- a/compressor/storage.py
+++ b/compressor/storage.py
@@ -2,6 +2,7 @@ import gzip
 import os
 from datetime import datetime
 import time
+from threading import Lock
 
 from django.core.files.storage import FileSystemStorage, get_storage_class
 from django.utils.functional import LazyObject, SimpleLazyObject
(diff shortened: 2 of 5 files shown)


diox:
Apologies for not getting back to you sooner on this, I have a shitty month and not much motivation for looking at code outside of work.

This PR works for offline compression, and we could go that route, but maybe we should try to see if we can address the broader issue first. I've written about it in I've commented about the broader issue in https://github.com/django-compressor/django-compressor/issues/1103#issuecomment-1055252728 : the heart of the problem is that compressor really cares about files being named in a predictable way, and to achieve that, it circumvents Django's storage implementation that doesn't let you do that by deleting the file first if it exists.

There are multiple problematic real-world scenarios with this:
- The one we're having in offline compression, where multiple threads are trying to write to the same file at the same time
- "Online" compression where 2 compressor threads try to write to the same file at the same time
- Tests running with multiple processes where 2 processes try to write to the same file at the same time
- "Online" compression where compressor tries to write to an existing file (therefore temporarily deleting it) while the file is being served to the browser by a completely different process

Maybe we could expand the lock idea to compressor core (adding it to the base storage class), to circumvent some of these scenarios ? It's a little scary to add locks to online compression, but maybe that should be ok with a short timeout just in case?

cuu508:
> Maybe we could expand the lock idea to compressor core (adding it to the base storage class), to circumvent some of these scenarios ? 

I see `django.core.files.storage` uses helper functions from `django.core.files.locks`. It looks like these functions lock access to files at OS level, and so would probably work even with multiple processes. So that's perhaps a piece of the puzzle.

Another idea: 

* drop the deletion logic in  `CompressorFileStorage.get_available_name()`
* this means that `get_available_name` will sometimes return filenames with an underscore and a random string appended at the end to ensure uniqueness
* override the `save()` method to add one additional last step: if the filename contains an underscore then move it into the correct location. 

So, instead of deleting a file and then writing new data in the same filename, compressor would write the data in a temporary file, then move it to the correct place. Python docs say `os.replace` is an atomic operation, so this would avoid the brief moment of file not existing between delete and write.

diox:
I like that idea of using `os.replace`. It seems a lot better than deleting the file if it exists! Compared to your original approaches, it might be a bit wasteful as we're writing the same file multiple times, but that seems fine. And we don't have to care about the

diox:
https://github.com/django-compressor/django-compressor/pull/1107 addressed this differently.

PR #1107, opened 2022-04-02
Remove delete() call from CompressorFileStorage.get_available_name

### LINKED ISSUE #1099: Intermittent failure in offline compression

cuu508:
This PR is an alternative to #1105 and #1100, and intends to fix #1099. It changes CompressorFileStorage:

* drops the deletion logic in  `CompressorFileStorage.get_available_name()`
* this means that `get_available_name` will sometimes return filenames with an underscore and a random string appended at the end to ensure uniqueness
* overrides the `save()` method to add one additional last step: if the filename contains an underscore then move it into the correct location. 

So, instead of deleting a file and then writing new data in the same filename, compressor would write the data in a temporary file, then move it to the correct place. Python docs say `os.replace` is an atomic operation, so this would avoid the brief moment of file not existing between delete and write.

Notes:

* I didn't add any new test cases. In theory there could be a test case which creates compressed files with clashing filenames, and then tests if the produced files contain any underscores (they shouldn't). But not sure if it's worth testing that.
* `GzipCompressorFileStorage` and `BrotliCompressorFileStorage` override the `save` method again and do some more file IO. I'm not sure if we could run into race conditions there too.

### DIFF
--- a/compressor/storage.py
+++ b/compressor/storage.py
@@ -33,13 +33,16 @@ class CompressorFileStorage(FileSystemStorage):
     def modified_time(self, name):
         return datetime.fromtimestamp(os.path.getmtime(self.path(name)))
 
-    def get_available_name(self, name, max_length=None):
-        """
-        Deletes the given file if it exists.
-        """
-        if self.exists(name):
-            self.delete(name)
-        return name
+    def save(self, filename, content):
+        temp_filename = super().save(filename, content)
+        # If a file already exists  in the target location, FileSystemStorage
+        # will generate an unique filename and save content there instead.
+        # When that happens, we move the file to the intended location using
+        # os.replace() (which is an atomic operation):
+        if temp_filename != filename:
+            os.replace(self.path(temp_filename), self.path(filename))
+
+        return filename
 
 
 compressor_file_storage = SimpleLazyObject(
--- a/compressor/tests/test_storages.py
+++ b/compressor/tests/test_storages.py
@@ -61,3 +61,9 @@ class StorageTestCase(TestCase):
         context = {'STATIC_URL': settings.COMPRESS_URL}
         out = css_tag("/static/CACHE/css/output.e701f86c6430.css")
         self.assertEqual(out, render(template, context))
+
+    def test_duplicate_save_overwrites_same_file(self):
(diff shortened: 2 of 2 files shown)


compressor/storage.py
```diff
@@ -33,13 +33,12 @@ def created_time(self, name):
...
-    def get_available_name(self, name, max_length=None):
-        """
-        Deletes the given file if it exists.
-        """
-        if self.exists(name):
-            self.delete(name)
-        return name
+    def save(self, filename, content):
+        temp_filename = super().save(filename, content)
+        if "_" in temp_filename:
```
diox, line 38: Why not compare `temp_filename` with `filename` instead ?

cuu508: Ah, of course! Thanks – fixed!

diox:
> I didn't add any new test cases. In theory there could be a test case which creates compressed files with clashing filenames, and then tests if the produced files contain any underscores (they shouldn't). But not sure if it's worth testing that.

I think it's worth testing (I know the current `CompressorFileStorage` implementation doesn't have any tests - we should change that). Something along the lines of what you suggest should be enough. Let me know if you don't have the motivation to do that (you already done a lot of work, I appreciate the time spent on this!) and I'll step in if needed.

> `GzipCompressorFileStorage` and `BrotliCompressorFileStorage` override the save method again and do some more file IO. I'm not sure if we could run into race conditions there too.

They do this with the low-level primitives directly though, so they might run into different issues. I think it's fine to ignore them for this.

cuu508:
No problem – added a small testcase which saves the same file and tests if the filename returned by `save()` contains any underscores or not.

diox (approved):
:+1: let's merge and get some people experiencing issues with this to test it!

PR #1110, opened 2022-05-10
Update django caching documentation link

tannart:
Updated a link pointing to the docs for the now deprecated django 2.1 to the docs for the dev edition of django.

### DIFF
--- a/docs/usage.txt
+++ b/docs/usage.txt
@@ -131,7 +131,7 @@ exception. If DEBUG is ``False`` these files will be silently stripped.
     Django's `caching documentation`_).
 
 .. _memcached: http://memcached.org/
-.. _caching documentation: https://docs.djangoproject.com/en/2.1/topics/cache/#memcached
+.. _caching documentation: https://docs.djangoproject.com/en/stable/topics/cache/#memcached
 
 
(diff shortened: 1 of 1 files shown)


docs/usage.txt
```diff
@@ -131,7 +131,7 @@ exception. If DEBUG is ``False`` these files will be silently stripped.
     Django's `caching documentation`_).
 
 .. _memcached: http://memcached.org/
-.. _caching documentation: https://docs.djangoproject.com/en/2.1/topics/cache/#memcached
+.. _caching documentation: https://docs.djangoproject.com/en/dev/topics/cache/#memcached
```
diox, line 134: Looks like /stable/ exists and redirects, it should be a better target.

```suggestion
.. _caching documentation: https://docs.djangoproject.com/en/stable/topics/cache/#memcached
```

diox:
Thanks for catching that!

docs/usage.txt
tannart: agreed, I was unaware of the existence of /stable/, in which case the link to template loaders on the same page (also points to dev) should be updated too.

diox (approved):
Great, thanks!

PR #1145, opened 2022-09-30
Update doc in light of new mangled name avoidance

kmtracey:
https://github.com/django-compressor/django-compressor/pull/1107 moved file-name-mangling avoidance into a `save` override, but the doc suggests calling `_save` which bypassess the new file-name-mangling-avoidance code. Updating a project to use django-compressor 4.0 that had a storages class which followed this doc resulted in mangled-named files appearing in the file system (and ultimately css changes not getting built/deployed, since the changes were in the mangled-name files while the rest of the process was using the "bare" name).

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -4,6 +4,11 @@ Changelog
 Unreleased
 ----------
 
+- Update the `CachedS3Boto3Storage` example storage subclass in "Remote Storages"
+  to work properly after the v4.0 change to how duplicate file names are handled
+  by `CompressorFileStorage`
+  
+
 v4.1 (2022-08-03)
 -----------------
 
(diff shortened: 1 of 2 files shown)


diox:
Thanks for catching this ! Maybe it's worth highlighting in the changelog as well ? (we already have an item talking about changes to duplicate filenames handling, but we could explicitly mention this caveat)

kmtracey:
Yes I think it would help to mention in changelog too, I wasn't sure if an update to an existing change log entry would be OK and/or if it should be noted on its own as a doc update related to the name mangling change?

diox:
I think we could do both. I don't mind altering the changelog, folks that read it through the package won't see it, but those reading it through readthedocs after we change it will benefit from it, it's better than nothing. 

Doesn't have to be a big thing, just a quick note/warning should be ok. If you're up for it feel free to make another commit in this PR for that, otherwise we can merge this change and I'll get to do more docs patching eventually.

diox (approved):
Perfect, thanks!

PR #1152, opened 2022-10-31
Simplify test settings file

diox:


diox:
Ah, can't be done in Django 3.2 :( Let's skip this for now.

PR #1161, opened 2022-11-30
Fix example for COMPRESS_OFFLINE_MANIFEST_STORAGE in docs

th3hamm0r:
Just run into my own faulty docs submitted with #1113.

#1160 is a bit related, as it also improves the docs around the new offline manifest storage.


diox (approved):
ah, good catch!

Issue #1182, opened 2023-04-06
Add support for django 4.2


foarsitter:
The test-suite needs to be run against 4.2 and main branch to verify compatibility. After confirming it is all working the classifiers can be updated.

jvacek:
https://github.com/django-compressor/django-compressor/pull/1189

PR #1189, opened 2023-05-13
Django 4.2 tox update

jvacek:
Fixes https://github.com/django-compressor/django-compressor/issues/1182


diox:
Thanks! You also need to edit the [GitHub Actions configuration](https://github.com/django-compressor/django-compressor/blob/develop/.github/workflows/ci.yml) for that change to do anything useful.

jvacek:
@diox thanks for the quick catch! patched above

jvacek:
Looks like all is passing @diox

PR #1205, opened 2023-10-09
Upgrade rcssmin == 1.1.2

areski:
There is a new version of rcssmin

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -164,7 +164,7 @@ setup(
     zip_safe=False,
     install_requires=[
         "django-appconf >= 1.0.3",
-        "rcssmin == 1.1.1",
+        "rcssmin == 1.1.2",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -164,7 +164,7 @@ def find_package_data(
     zip_safe=False,
     install_requires=[
         "django-appconf >= 1.0.3",
-        "rcssmin == 1.1.1",
+        "rcssmin == 1.1.2",
         "rjsmin == 1.2.1",
```
jonpas:
```suggestion
        "rjsmin == 1.2.2",
```
Also new version of `rjsmin`, might as well bump both.

areski:
Thanks @jonpas

diox (approved):

PR #1212, opened 2023-11-23
Add Read The Docs configuration file.

tim-schilling:
Copied from https://docs.readthedocs.io/en/stable/config-file/v2.html

Switched to Python 3.11, since the project doesn't explicitly use 3.12 yet.


carltongibson (approved):
Brilliant! You can guess how happy I am you picked this up! 

Thank you all 🚀

PR #1226, opened 2024-01-17
Keep GitHub Actions up to date with Dependabot

cclauss:
https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,4 +8,4 @@ updates:
 - package-ecosystem: github-actions
   directory: "/"
   schedule:
-    interval: daily
+    interval: weekly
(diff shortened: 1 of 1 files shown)


carltongibson (commented):
I wonder if a weekly update isn't enough? 

I also saw something about grouping updates... 🤔 Do you know about that? (Sometimes I feel dependabot can be a bit noisy...)

cclauss:
Switched to `weekly`.

This repo only has just two Actions so I doubt that the volume of updates will be overwhelming.
https://github.com/django-compressor/django-compressor/blob/7face0f5c3b2c6babfb777cd962979d58730550a/.github/workflows/ci.yml#L57-L58

carltongibson (approved):
Yes, this looks good to me. Thanks. 

(It was the pip packages that could do with grouping I think. Separate PR) 

Thanks.

Issue #1101, opened 2022-02-23
Drop slimit


diox:
https://github.com/django-compressor/django-compressor/runs/4690546430?check_suite_focus=true

Only 3.6 builds seem to be passing, rest are failing with:

```
======================================================================
ERROR: test_slimit_filter (compressor.tests.test_filters.SlimItTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/django-compressor/django-compressor/compressor/filters/base.py", line 84, in __init__
    func = getattr(import_module(mod_name), func_name)
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/runner/work/django-compressor/django-compressor/.tox/py37-2.2.X/lib/python3.7/site-packages/slimit/__init__.py", line 27, in <module>
    from minifier import minify
ModuleNotFoundError: No module named 'minifier'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/runner/work/django-compressor/django-compressor/compressor/tests/test_filters.py", line 210, in test_slimit_filter
    self.assertEqual(output, SlimItFilter(content).output())
  File "/home/runner/work/django-compressor/django-compressor/compressor/filters/base.py", line 96, in __init__
    "correctly installed." % (self.callback, warning))
django.core.exceptions.ImproperlyConfigured: The callback slimit.minify couldn't be imported. Make sure the dependency (slimit) is correctly installed.
```

diox:
slimit hasn't had a release for a very long time. If people want to keep using it they can easily write the 3 lines to do that themselves. I'm going to deprecate it and remove the tests in the meantime.

PR #1309, opened 2025-10-12
Add Python 3.13 to CI and declare support

browniebroke:
- Add tox targets and CI jobs following https://docs.djangoproject.com/en/5.2/faq/install/#what-python-version-can-i-use-with-django
- Add trove classifiers

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,8 @@ jobs:
             toxenv: py311-5.1.X
           - python: "3.12"
             toxenv: py312-5.1.X
+          - python: "3.13"
+            toxenv: py313-5.1.X
 
           - python: "3.10"
             toxenv: py310-5.2.X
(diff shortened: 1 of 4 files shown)


browniebroke:
The tests fail to install the dependencies, it seems to be because `slimit` requires lib2to3, which was removed in Python 3.13:

https://github.com/rspivak/slimit/blob/3533eba9ad5b39f3a015ae6269670022ab310847/setup.py#L27-L31

It seems that [django-compressor considered dropping this dependency](https://github.com/django-compressor/django-compressor/issues/1101) at some point and is now deprecated. Might be time to actually drop it.

browniebroke:
Actually, the `SlimItFilter` was removed in https://github.com/django-compressor/django-compressor/pull/1150 so we don't need to install the dependency anymore.

browniebroke:
I tried to also add 3.14 but it seems that will need more work: https://github.com/django-compressor/django-compressor/actions/runs/18445151249/job/52550802176

carltongibson (approved):
Great. Thanks.

PR #1311, opened 2025-10-12
Drop support for Python 3.9

browniebroke:
Will be EOL at the end of the month


carltongibson (approved):
Yep. Great. It's time. Thanks!

PR #1312, opened 2025-10-12
Remove version pinning for rcssmin and rjsmin

### LINKED ISSUE #1195: Dependencies are pinned
Hi! I package this project for Arch Linux.

For system packaging it is problematic to pin dependencies the way it is done in this project:

https://github.com/django-compressor/django-compressor/blob/939499ecca8886b3192da1343f20b30ab248baa4/setup.py#L167-L168

Downstream we have to remove the specific version pinning, as we are a floating target where only one version of a given library can exist. However, even for user installs of this library it is problematic, as installing django-compressor requires a specific version of these libraries and a different library may require them to be less strict.

browniebroke:
Fix #1195

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -166,7 +166,7 @@ setup(
     install_requires=[
         "Django >= 4.2",
         "django-appconf >= 1.0.3",
-        "rcssmin == 1.2.1",
-        "rjsmin == 1.2.4",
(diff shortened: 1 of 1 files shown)


carltongibson (approved):
Yes, super. Thank you.

PR #1313, opened 2025-10-12
Migrate packaging to pyproject.toml

browniebroke:


.github/workflows/ci.yml
```diff
@@ -51,7 +51,6 @@ jobs:
           python-version: ${{ matrix.versions.python }}
       - run: |
           pip install tox setuptools
```
carltongibson: Do we still need setuptools here?

carltongibson (commented):
Looks great. Just one question...

.github/workflows/ci.yml
browniebroke: Good point, I don't think so

browniebroke:
```suggestion
          pip install tox
```

carltongibson (approved):
Lovely!

PR #1324, opened 2026-01-21
Add Subresource Integrity (SRI) support for compressed assets

anuj9196:
- Introduced COMPRESS_SRI_HASHES and COMPRESS_SRI_CROSSORIGIN settings for configuring SRI.
- Updated Compressor class to generate integrity attributes for <script> and <link> tags.
- Modified templates to include integrity and crossorigin attributes when applicable.
- Added tests for SRI functionality in CSS and JS tags.
- Updated documentation to reflect new SRI settings and usage examples.

### DIFF
--- a/compressor/base.py
+++ b/compressor/base.py
@@ -1,5 +1,7 @@
 import os
 import codecs
+import base64
+import hashlib
 from importlib import import_module
 from urllib.request import url2pathname
 
@@ -392,7 +394,14 @@ class Compressor:
         if not self.storage.exists(new_filepath) or forced:
             self.storage.save(new_filepath, ContentFile(content.encode(self.charset)))
         url = mark_safe(self.storage.url(new_filepath))
-        return self.render_output(mode, {"url": url})
+        context = {"url": url}
+        integrity = self.get_integrity(content)
+        if integrity:
+            context["integrity"] = integrity
+            crossorigin = settings.COMPRESS_SRI_CROSSORIGIN
+            if crossorigin:
+                context["crossorigin"] = crossorigin
+        return self.render_output(mode, context)
 
     def output_inline(self, mode, content, forced=False, basename=None):
         """
@@ -435,3 +444,18 @@ class Compressor:
         )
         template_name = self.get_template_name(mode)
         return render_to_string(template_name, context=final_context)
+
+    def get_integrity(self, content):
+        """
+        Returns the Subresource Integrity (SRI) string for the given content.
+        """
+        hashes = settings.COMPRESS_SRI_HASHES
+        if not hashes:
+            return ""
+        content_bytes = content.encode(self.charset)
+        integrity_parts = []
+        for algo in hashes:
+            digest = hashlib.new(algo, content_bytes).digest()
+            encoded = base64.b64encode(digest).decode("ascii")
+            integrity_parts.append("%s-%s" % (algo, encoded))
+        return " ".join(integrity_parts)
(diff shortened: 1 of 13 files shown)


compressor/tests/test_templatetags.py
```diff
@@ -24,6 +26,12 @@ def render(template_string, context_dict=None, context=None):
     return t.render(c).strip()
 
 
+def build_integrity(compressor):
```
diox: Tests are not passing and would need to be fixed, but please remove this function and hardcode expected integrity values instead where we want them - otherwise if `compressor.get_integrity()` is broken internally tests would still pass.

compressor/base.py
```diff
@@ -435,3 +444,18 @@ def render_output(self, mode, context=None):
         )
         template_name = self.get_template_name(mode)
         return render_to_string(template_name, context=final_context)
+
+    def get_integrity(self, content):
```
diox, line 454: This should ideally have its own unit test

compressor/conf.py
```diff
@@ -147,3 +153,29 @@ def configure_precompilers(self, value):
                 "missing commas."
             )
         return value
+
+    def configure_sri_hashes(self, value):
```
diox, line 157: This should also have its own tests

diox (changes requested):
Thanks for the PR, we're close! Let's improve tests and make them pass.

anuj9196:
@diox pull request updated with the unit tests. Let me know if any further change is required

compressor/cache.py
```diff
@@ -97,7 +97,12 @@ def write_offline_manifest(manifest):
 
 
 def get_templatetag_cachekey(compressor, mode, kind):
-    return get_cachekey("templatetag.%s.%s.%s" % (compressor.cachekey, mode, kind))
+    sri_hashes = ",".join(settings.COMPRESS_SRI_HASHES or ())
+    crossorigin = settings.COMPRESS_SRI_CROSSORIGIN or ""
```
diox: That logic should live in the `cachekey` property of the `Compressor` class (`compressor/base.py` around line 240), since the class is responsible for the output being different depending on those settings. I don't think we need to modify `get_templatetag_cachekey()` at all.

anuj9196: Test was failing without it. You can check checks on previous commits. Modifying this method fixed the test run.

Please check again and let me know.

diox: I realize we do need something in the cache key, I'm just saying this is the wrong place for it, since the logic around integrity is in the `Compressor` class.

Instead of modifying `get_templatetag_cachekey()` I'm suggesting to do something like:
```diff
diff --git a/compressor/base.py b/compressor/base.py
index 58a8fd9..280c16a 100644
--- a/compressor/base.py
+++ b/compressor/base.py
@@ -239,7 +239,13 @@ class Compressor:
     @cached_property
     def cachekey(self):
         return get_hexdigest(
-            "".join([self.content] + self.mtimes).encode(self.charset), 12
+            "".join(
+                [self.content]
+                + self.mtimes
+                + list(settings.COMPRESS_SRI_HASHES)
+                + [settings.COMPRESS_SRI_CROSSORIGIN or ""]
+            ).encode(self.charset),
+            12,
         )
 
     def hunks(self, forced=False):
```

anuj9196: @diox Thanks for the explanation. I have update the PR as requested.

diox (approved):
Thanks for the PR!

PR #1325, opened 2026-02-13
Fix typo in OFFLINE_MANIFEST_STORAGE_ALIAS documentation and code

nstuardod:
Correct a typo in `OFFLINE_MANIFEST_STORAGE_ALIAS` within both the documentation and the code.

This typo was causing some projects to not copy the manifest to the correct storage.


carltongibson (approved):
Yes. Good catch. Thanks.

